PR #4, opened 2016-04-07
[docker]Add python-mysqldb dependency for old scripts

sanacl:


docker/gelk/Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN apt-get update && \
         python3-requests python3-dateutil python3-bs4 \
         python3-pip python3-dev python3-redis python3-sqlalchemy \
         python3-flask \
+	python-mysqldb \
```
acs: Could you use spaces instead of real tabs? Thanks!!!

acs:
Thanks lcanas

PR #6, opened 2016-05-05
[Cauldron dashes] Changed time from to now-1y

sanacl:


acs:
Great @sanacl but this files are generated from live dashboards, so next time we update the dashboards those changes will be lost. It is better to define the 1y period in the dashboards in kibana which are exported and used as templates.

acs:
After talking with @sanacl we decided to merge this PR and that the process for working with dashboard starts importing the dashboards in the git repo.

PR #7, opened 2016-05-06
[Cauldron dashes] New cauldron dashes

sanacl:


acs:
Maybe I will move experimental dashes to other directory than attic but it is ok!

acs:
@sanacl have you removed the experimental dashes from the original place? It seems that no!

PR #15, opened 2016-07-06
[mordred] Change usage projectinfo.py

zhquan:


mordred/utils/projectinfo.py
```diff
@@ -32,11 +32,60 @@
 
 REPOSITORIES_AVALIABLE = ['source_repo', 'github', 'mailing_lists', 'telegram', 'kuma',
                           'kitsune', 'mozilla_reps', 'mediawiki', 'irc', 'confluence',
-                          'jira',' maniphest', 'gerrit']
+                          'jira',' maniphest', 'gerrit', 'meetup']
+
+
+def add_subparser(subparsers):
```
sduenas: IMHO, it is not a good practice to pass an object as argument and modify it there. There are better ways to do this.

mordred/utils/projectinfo.py
```diff
@@ -32,11 +32,60 @@
...
+    """Add subparser"""
+
+    add_usage = "projectinfo.py file add [--help] project_name [--repo <repo_type> <url>]"
+    add_parser = subparsers.add_parser("add",
+                                    help="add project",
+                                    usage=add_usage)
+    add_parser.add_argument("project_name",
+                        action="store",
```
sduenas: Please check the indentation in the whole file!

mordred/utils/projectinfo.py
```diff
@@ -32,11 +32,60 @@
...
 REPOSITORIES_AVALIABLE = ['source_repo', 'github', 'mailing_lists', 'telegram', 'kuma',
                           'kitsune', 'mozilla_reps', 'mediawiki', 'irc', 'confluence',
-                          'jira',' maniphest', 'gerrit']
+                          'jira',' maniphest', 'gerrit', 'meetup']
+
+
+def add_subparser(subparsers):
+    """Add subparser"""
+
+    add_usage = "projectinfo.py file add [--help] project_name [--repo <repo_type> <url>]"
```
sduenas: You don't need to add this message because `argparser`library can do it for you.

mordred/utils/projectinfo.py
```diff
@@ -50,49 +99,26 @@ def read_arguments():
...
+    add_subparser(subparsers)
+    rm_subparser(subparsers)
+    list_subparser(subparsers)
 
     args = parser.parse_args()
 
-    if args.repo:
-        if args.repo not in REPOSITORIES_AVALIABLE:
+    if args.command != "list" and args.repo:
+        if args.repo[0] not in REPOSITORIES_AVALIABLE:
```
sduenas: Why do you need a list for this argument?

mordred/utils/projectinfo.py
```diff
@@ -239,23 +257,19 @@ def run(conf):
...
-        conf['repo'] = args.repo
-    if args.url:
-        conf['url'] = args.url
+
+    if args.command == "list":
+        if args.parent:
+            conf['parent'] = args.parent
+    else:
+        if args.repo:
+            conf['repo'] = args.repo[0]
```
sduenas: What happens when `repo` has only one value or none?

sanacl:
I spotted some bugs, for instance I can not add a parent project because the option is not added to the parser. In any case this is becoming more urgent. I'm accepting the patch and iterating over it. As soon as we have a running version, we can be more strict. Hope you agree @sduenas

sduenas:
In my opinion, if there are known bugs a patch cannot be accepted. I don't mind whether it's urgent or not. We need to do the things right, no matter how urgent is. If we accept anything people get lazy and don't push themselves to improve their code nor learn anything new. My two cents.

PR #40, opened 2016-08-02
[mordred] Added support for twitter

dpose:
Added support for twitter

### DIFF
--- a/mordred/bin/mordred.sh
+++ b/mordred/bin/mordred.sh
@@ -36,6 +36,10 @@ function set_variables {
     DB_SH=$PROJECT_SHORTNAME"_sh"
     DB_PRO=$PROJECT_SHORTNAME"_pro"
 
+    if [ -z TWITTER_ENRICHED_INDEX ]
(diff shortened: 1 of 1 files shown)


mordred/bin/mordred.sh
```diff
@@ -261,6 +267,13 @@ function retrieve_data {
         pipermail_retrieval
         log_result "[pipermail - mls] Retrieval finished" "[pipermail - mls] ERROR: Something went wrong with the retrieval"
     fi
+
+    if [ $TWITTER_ENABLED -eq 1 ]
+        then
+        log "[twitter] Retrieving data"
```
sanacl: I prefer not to say "Retrieving data" when no data is retrieved. Disable these two messages.

dpose:
Ok. I added these lines to avoid thinking there is a problem with the data collection when the log is seen. But it can be confusing too, so I'll delete them.

dpose:
I've already deleted those two messages.

PR #51, opened 2016-09-19
[mordred] Fix issue when SH is called

dpose:
Parameter "-h" in SH is used for help not for specify the host.
$DB_HOST is duplicated in some calls to SH.

### DIFF
--- a/mordred/bin/mordred.sh
+++ b/mordred/bin/mordred.sh
@@ -618,7 +618,7 @@ function git_enrichment {
 
     REPOS=`get_repo_list source_repo`
 
-    if [ $ENR_EXTRA_FLAG == '--only-studies' ]; then
+    if [ "$ENR_EXTRA_FLAG" == "--only-studies" ]; then
 	# when we execute studies, we want it without going repo by repo
         REPOS="''"
     fi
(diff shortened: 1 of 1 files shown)


mordred/bin/mordred.sh
```diff
@@ -618,7 +618,7 @@ function git_enrichment {
 
     REPOS=`get_repo_list source_repo`
 
-    if [ $ENR_EXTRA_FLAG == '--only-studies' ]; then
+    if [ "$ENR_EXTRA_FLAG" == "--only-studies" ]; then
```
sanacl: This is incorrect

mordred/bin/mordred.sh
```diff
@@ -764,10 +764,10 @@ function group_unaffiliated_people {
         log "[sortinghat] Affiliating the unaffiliated people to group $SH_UNAFFILIATED_GROUP"
 
         if [ -z $DB_PASSWORD ]; then
-            sortinghat --host $DB_HOST -d $DB_SH -u $DB_USER -h $DB_HOST orgs -a $SH_UNAFFILIATED_GROUP >> $LOGS_DIR"/sortinghat.log" 2>&1
+            sortinghat --host $DB_HOST -d $DB_SH -u $DB_USER orgs -a $SH_UNAFFILIATED_GROUP >> $LOGS_DIR"/sortinghat.log" 2>&1
```
sanacl: This is also incorrect, we need to specify the DB_HOST.

sanacl:
You were right in the notification of the bug (not in the solution). What we need is a new PR which change -h with --host

PR #52, opened 2016-09-23
[mordred] Updated the support for gerrit

dpose:
Updated the support for gerrit according to the
commit '5eb165ae65cd3dd685cd3e650411a119446b253d'
of perceval.


albertinisg:
I've spoke with @dpose and he told me it's working in a pre-production environment, and he needs it straight away. Accordingly with https://github.com/grimoirelab/perceval/commit/5eb165ae65cd3dd685cd3e650411a119446b253d, the `url` was set as a positional argument, so this change makes all sense to me.

sanacl:
Great, thank you guys.

PR #54, opened 2016-09-29
[google_hits] Variable hits converted to integer before storing

sanacl:


acs:
This was fixed in a change in my local repository. Sorry for not uploading it yet! 

https://github.com/grimoirelab/GrimoireELK/commit/2460aa25c7325079a28fa161904ffa04e7115389

PR #55, opened 2016-10-03
[mordred] Fix wrong p2o call to collect gerrit data

sanacl:


sanacl:
@dpose just told it is ok.

PR #59, opened 2016-10-05
[mordred] Add script to generate new JSON format files for project

canasdiaz:
Add script to generate the two JSON files we use now both for projects hierarchy and project-repos definition.

### DIFF
--- a/mordred/utils/json/eclipse2bitergia.py
+++ b/mordred/utils/json/eclipse2bitergia.py
@@ -1,0 +1,157 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2016 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+#
+# Authors:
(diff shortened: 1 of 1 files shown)


sanacl:
Guys ( @acs, @dpose ), I would like both of you to review the code. Thank you!

jgbarah (commented):
Instead of options designed in a way that you must run the script twice for doing the most usual think that the script is going to do (producing the two files), you better design it in a way that you can run it at once. A little tweak, for example, could help. Consider this shell command interface:

```
eclipse2bitergia.py -f eclipse.json -h hierarchy.json -p projects.json
```

If you only need one of them (-h or -p), just include only one of them.

This way, operation is simpler (you only need to call it once), and performance too (if reading the eclipse.json file takes a while, because it is long, you only need to read it once).

Besides, I didn't look at the source code, so maybe you already did this. But it is better if you can build it so that there is a function that mimics the shell command API, and that's the only thing that the main program does. That way, if in the future you want to integrate the code in a script, you just call that function instead of the command line.

In this specific case, since the function is dealing with JSON documents, there could be two levels: one with exactly the command line API (a function using file names as parameters), and a lower level accepting Python dictionaries as arguments, so that conversion to/from JSON (or YAML, or whatever) can be done elsewhere.

sanacl:
Changed.

```
luis@hogaza ~/repos/my-forks/GrimoireELK/mordred/utils/json ±new_json_formatters⚡ » python3 eclipse2bitergia.py -h
usage: eclipse2bitergia.py -f <input_file> -y <hierarchy_file> -p <projects_file>

optional arguments:
  -h, --help            show this help message and exit
  -f INPUT_FILE, --input_file INPUT_FILE
                        Input file
  -y HIERARCHY_FILE, --hierarchy HIERARCHY_FILE
                        Output file for hierarchy
  -p PROJECTS_FILE, --projects PROJECTS_FILE
                        Output file for projects

```

jgbarah (approved):
Looks ok to me now!

dpose:
It looks good to me too.
I'll test it as soon as I can to give more feedback.

sanacl:
@acs ??

acs:
@sanacl I am a bit out if context in this script. It is pretty simple so if it is ok for you, it is good for me also. For me it is a bit strange to generate several files at the same time using command line params, but if the goal of this script is that, it is ok.

PR #64, opened 2016-11-03
[mordred] Fix supybot parsing for launching p2o

albertinisg:
There were an issue while launching p2o, the supybots elements were not being parsed correctly making the tool fail


dpose:
I've checked it works successfully after making these changes, so I've merged this PR.

PR #67, opened 2016-11-11
[mordred] Add support for fetching the whole JIRA website

albertinisg:
In 6c3c6d4fedc19cacb366f103b2f78e8e4796798f I've introduced support for JIRA projects, but with that approach I broke the existing one of fetching the whole JIRA website. Before, there was a variable in mordred checking the website, now it should be added in the JSON file like other data sources.

BTW I already did test it in pre-pro environment


dpose:
LGTM.

PR #72, opened 2016-12-29
Modify import clauses to do it as much relative as possible.

jgbarah:
In many places we had absolute module names, such as grimoire.elk.enrich.
This patch makes them relative, such as .enrich, if the importing
file is in grimoire/elk.

This is expected to easy the transtion to a new module name for the library.


jgbarah:
Added another commit that changes the name of the module to grimoire_elk, and the name of the generated package to grimoire-elk. See more details in the commit message.

Both commits complete the transition to the new name, plus have everything needed for producing the new package. In fact, there is already a package, installable with pip, that includes these two commits. Try it with

```
pip install grimoire-elk
```

acs:
@jgbarah could you update the PR so it has no conflicts with master? Sorry about the last modifications that breaks your PR, but they are needed in production now. Thanks!

jgbarah:
Done!

PR #74, opened 2017-01-28
Add dependency on perceval-mozilla

jgbarah:
perceval-mozilla is needed fpr p2o to run. For now, I'm including
it as a dependency. Probably we should make grimoire_elk a bit
smarter, and use perceval-mozilla only if it is available, or
else, throw an exception if it is invoked when calling the module,
but it is not installed.

I bumped version to 0.22.1 to upload a new package to pypi.


acs:
Yes, right now all the perceval backends are loaded in gelk, no matter they will be used later or not. We need to do this in an smarter way (just importing the needed backends). @jgb could you open an issue for tracking it? Thanks!

PR #75, opened 2017-01-29
Add support for grimoire-kidash pip package.

jgbarah:
Changes to allow kidash.py to be pckaged as a separate package.
This allows for easy installation of kidash.py, assuming
grimoire-elk is already installed (otherwise, it will be pulled
by dependencies in grimoire-kidash).

### DIFF
--- a/README_kidash.md
+++ b/README_kidash.md
@@ -1,0 +1,8 @@
+# Kidash
+
+Kidash is a prototype of a tool for managing Kibana dashboards from the command line. It is a part of [GrimoireLab](https://grimoirelab.github.io).
+
+## Source code
+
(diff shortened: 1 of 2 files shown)


acs:
@jgbarah after more testing, when I try to use it:

```
(acs@dellx) (master $% u=) ~/devel/GrimoireELK $ sudo python3 setup_kidash.py install
....
Warning: pypandoc module not found, or pandoc not installed. Using md instead of rst
error: [Errno 2] No such file or directory: '/tmp/easy_install-2e82jrlk/grimoire-elk-0.22.1/README.md
```

acs:
You need pandoc installed as package. This error is just a warning, so it should be ok.

PR #88, opened 2017-07-05
Remove deprecated Kibana panel JSON files

dicortazar:
The new repository for panels is available at
https://github.com/grimoirelab/panels


acs:
Hmmm, there are panels in this directory for Maniphest that are what we are using in our dogfooding. But we are not using it nowdays. @sanacl what do you think? Probably we can remove them and if needed, retrieve them from git history.

jgbarah:
Assigned to @sanacl so that he (hopefully) sees this...

dicortazar:
Comments?

acs:
@dicortazar we are at the same point. In this directory for example we have the panels we created for phabricator that we use to track our internal activity. We need to review what it is here and to decide what to do.

dicortazar:
Hey, why don't we migrate those panels to the panels repo? :)

And then remove those from here ;)...

acs:
Sure, but probably those panels will need some love before adding them to panels repository. But it is the way to go!

sanacl:
I would just remove the directory

dicortazar:
This PR does that.

acs:
LGTM. Ok, let's move all this stuff to attic! Thanks guys!

PR #89, opened 2017-07-05
Remove deprecated mordred source code

dicortazar:
The new Mordred tool is available at https://github.com/grimoirelab/mordred


acs:
@sanacl if you are ok with this removal let's do it! Thanks @dicortazar for cleaning the repository!

PR #91, opened 2017-07-15
Move kidash to a new directory.

jgbarah:
This way it can be more easily covnerted into a pip package.
I included some minor changes to the rest to match this change.


acs:
Ok, LGTM according to our conversation past Friday. So merging it. Thanks @jgbarah

PR #97, opened 2017-08-14
Fix string to integer concatenation error.

aghassabian:
got error   `unsupported operand type(s) for +: 'int' and 'str'` on gerrit ocean. fixed.


dpose (approved):
Thank you very much for your contribution @aghassabian

PR #106, opened 2017-10-09
Downgrade dependency on perceval-opnfv

jgbarah:
The latest release of perceval-opnfv is 0.1.0, so we better have that as dependency, otherwise there is no package to satisfy it ;-)


acs (approved):
LGTM

PR #110, opened 2017-10-19
[enrich] Use _version.py to get GELK version.

jgbarah:
Version was obtained by getting the git commmit description,
using the git command. Instead of that, use the _version.py file
that we maintain for consistent versioning.

The previous schema, based on the git command,  does not work
with pip packages, since they are not clones of the git repo.


jgbarah:
Any news on this?

jgbarah:
I found a simpler way of referencing the version: __version__ in main module. Using it now (new commit with that code pushed, it is the one currently for review).

PR #113, opened 2017-10-31
[kidash] Add support for using panels in the panels module

jgbarah:
Now, panels (JSON files with panel defintions) are looked for
first in a local directory panels/json, and if not found, in the
panels module, if it is present and importable.
For maintaining compatibility, panels with names such as
"panels/json/git.json" are supported. But the most simple
"git.json" is also supported, meaning "git.json" file in the
current directory, or "git.json" file in the corresponding
directory of the modules file.

See README.md for more details.

Versions for both grimoire-elk and grimoire-kidash are updated.

grimoirelab-panels is now a dependency of grimoire-kidash.

### DIFF
--- a/grimoire_elk/panels.py
+++ b/grimoire_elk/panels.py
@@ -26,11 +26,13 @@
 import json
 import logging
 
-from os import sys
+import os
+import os.path
+import pkgutil
+import sys
 
 from .elk.elastic import ElasticSearch
 from .elk.utils import grimoire_con
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


jgb:
@acs I'll ask it one more time: please use the correct alias @jgbarah !

jgbarah:
@acs, for example: I want to upload the standard git dashboard to Kibiter. I would only need:

```
$ pip install grimoire-kidash
$ kidash.py --elastic_url-enrich http://localhost:9200 --import git.json
```

No need to clone the `panels` repo, no need to even install the `grimoirelab-panels` package (it is pulled as a dependency by `grimoire-kidash`.

For the `grimoirelab-panels` to be really useful, it is important that the user does not need to know where the JSON for the panels are, but that they are available as much transparently as possible. 

Something similar happens with `mordred`, which also pulls `grimoirelab-panels` as a dependency. Now, it is possible to write in a mordred configuration file:

```
  menu:
  - name: Overview
    panel: git.json
  - name: Attraction/Retention
    panel: git_demographics.json
```

instead of:

```
  menu:
  - name: Overview
    panel: panels/json/git.json
  - name: Attraction/Retention
    panel: panels/json/git_demographics.json
```

and again, no need to have the `panels` repo cloned, or the package installed explicitly.

In addition, the current behavior is maintained, and has precedence, which means that if you prefer to refer to the panel as `panels/json/git.json`, and that file is available, that's fine, it will work (with independence of what is in the `grimoirelab-panels` package.

Even more, if you keep using `panels/json/git.json`, but you prefer not to clone the `panels` repo, that's fine as well, since that's managed as an special case, and the corresponding panel will be retrieved from the package.

Is it maybe more clear now?

acs:
Yes @jgbarah it is more clear now. But as a user of kidash, it will be a kind of "black magic" how git.json is found if it is not from a file. But it is clear that the feature is useful. Let's talk about it tomorrow!

acs:
@jgbarah after our meeting yesterday, let's merge this PR that also open the possibility to having a "marketplace" for panels that can be accessed from kidash easily. Let's evolve the idea!

grimoire_elk/panels.py
```diff
@@ -342,61 +344,90 @@ def list_dashboards(elastic_url, es_index=None):
...
-        if 'dashboard' not in kibana:
-            logger.error("Wrong file format. Can't find 'dashboard' field.")
-        name = kibana['dashboard']['id']
-
-    return name
+            import panels
+            logger.debug("Reading panel from module panels")
+            # Next is just a hack for files with "expected" prefix
+            if panel_file.startswith('panels/json/'):
+                module_file = panel_file[len('panels/json/'):]
```
acs: This two lines can be changed with:

panel_file = panel_file.replace('panels/json/', '')

but I will accept as it is now in your code

jgbarah:
Thx!!!!

PR #114, opened 2017-11-02
Git backend - update for Gitlab / new Gitlab backend

assadOW2:
Since at OW2 we are using Gitlab, I did some little changes in the original code in order to handle gitlab urls as well as github. I thought it might be interesting to handle gitlab as well since almost everything is the same as Github.
I didn't update the API login, hence the #TODO tag
I didn't neither change variable or function names, they are still "github" based.

### DIFF
--- a/grimoire_elk/ocean/gitlab.py
+++ b/grimoire_elk/ocean/gitlab.py
@@ -23,9 +23,8 @@
 #   Alvaro del Castillo San Felix <acs@bitergia.com>
 #
 
-from dateutil import parser
-
 from .elastic import ElasticOcean
+
 
 class GitLabOcean(ElasticOcean):
     """GitLab Ocean feeder"""


acs:
Hi @assadOW2!

Do you plan to add the API login feature? Without it I suppose supporting GitLab is not useful at all becasue the rate limit will be reached and the query for getting the commit information, for getting the username of the commit in GitLab, will fail.

assadOW2:
Hi @acs !

By API login, I was thinking about _get_github_login_ function, which is used only when using sortinghat.
As it is in my fork, you can get the informations you're mentioning such as username, commit, etc ...
To be honest, I haven't tried sortinghat yet.

assadOW2:
I have committed new files in order to add a new backend, Gitlab. It is based on Github backend but when dealing with issues, lots of metadata are different. So I decided to add a new one in order to make things cleaner.

acs:
@assadOW2, great, I have seen that you are doing it at https://github.com/grimoirelab/perceval/pull/195. Good luck! It is the place to do all GitLab data collection.

valeriocos:
Support for GitLab has been recently included to ELK (https://github.com/chaoss/grimoirelab-elk/pull/409), so I would close this PR. 
@assadOW2 if you want to improve the raw and/or enrich connectors, feel free to submit another PR

jgbarah:
Is this also available in Mordred?

PR #115, opened 2017-11-06
[setup] Improve setup.py

### LINKED ISSUE #101: Dependencies on sortinghat and MySQLdb not declared
Noticed this when running the tests in a build environment. They should be in install_requires or extras_requires

jgbarah:
Add some parameters to setup.py, including extras_require (for sortinghat and arthur), and fix minor issue with generation of README.rst.

Closes #101.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,5 +6,16 @@ GrimoireELK is a prototype of the *Grimoire Open Development Analytics platform*
 
 The following packages are produced from this repository:
 
-* `grimoire-elk`: [![PyPI version](https://badge.fury.io/py/grimoire-elk.svg)](https://badge.fury.io/py/grimoire-elk) 
+* `grimoire-elk`: [![PyPI version](https://badge.fury.io/py/grimoire-elk.svg)](https://badge.fury.io/py/grimoire-elk)
+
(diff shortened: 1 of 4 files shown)


jgbarah:
@sebix said:

> Some files depend on MySQLdb, which is still not required:

MySQLdb should be pulled in by SortingHat. Does it not work that way for you? (in fact, MySQLdb is only needed by SortingHat...)

sebix:
No. MySQLdb is a depedency of this package and should be declared as such.

Additionally, sorthinghat is only an optional dependency.

jgbarah:
I cannot figure out why grimoire-elk should ever use MySQLdb, except for dealing with SortingHat. I will have a look to check what's happening... Thanks for the heads up, @sebix.

sebix:
> I cannot figure out why grimoire-elk should ever use MySQLdb,

In grimoire_elk/elk/database.py you are using it. Whenever you import that module, you'd get an ImportError if you do not have that library.

jgbarah:
OK, for now, I've added a new extra dependency on `PyMySQL`, since `MySQLDB` doesn't work with Python3 (AFAIK). In any case, my impression is that we're not using SQL from `GrimoireELK` anymore, and that functionality is now in `mordred`. I will check, and maybe submit a pr later removing that, if I'm right.

PR #120, opened 2017-11-08
[jira] Add better checks when looking for sprints and story points

jgbarah:
The check we had when looking for sprints and story points failed in some cases. Improve it, so that it doesn't fail with other formats of custom fields.

### DIFF
--- a/grimoire_elk/elk/jira.py
+++ b/grimoire_elk/elk/jira.py
@@ -214,13 +214,15 @@ class JiraEnrich(Enrich):
 
         for field in issue['fields']:
             if field.startswith('customfield_'):
-                if issue['fields'][field]['name'] == "Story Points":
-                    eitem['story_points'] = issue['fields'][field]['value']
-                elif issue['fields'][field]['name'] == "Sprint":
-                    value = issue['fields'][field]['value']
-                    if value is not None and len(value)>0:
-                        sprint = value[0].partition(",name=")[2].split(',')[0]
-                        eitem['sprint'] = sprint
+                if type(issue['fields'][field]) is dict:
+                    if 'name' in issue['fields'][field]:
+                        if issue['fields'][field]['name'] == "Story Points":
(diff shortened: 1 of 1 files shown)


grimoire_elk/elk/jira.py
```diff
@@ -214,13 +214,16 @@ def get_rich_item(self, item):
...
             if field.startswith('customfield_'):
-                if issue['fields'][field]['name'] == "Story Points":
-                    eitem['story_points'] = issue['fields'][field]['value']
-                elif issue['fields'][field]['name'] == "Sprint":
-                    value = issue['fields'][field]['value']
-                    if value is not None and len(value)>0:
-                        sprint = value[0].partition(",name=")[2].split(',')[0]
-                        eitem['sprint'] = sprint
+                if type(issue['fields'][field]) is dict:
+                    print(str(issue['fields'][field]))
```
acs: Could you remove the print?

grimoire_elk/elk/jira.py
```diff
@@ -214,13 +214,16 @@ def get_rich_item(self, item):
...
+                    if 'name' in issue['fields'][field]:
+                        if issue['fields'][field]['name'] == "Story Points":
+                            eitem['story_points'] = issue['fields'][field]['value']
+                        elif issue['fields'][field]['name'] == "Sprint":
+                            value = issue['fields'][field]['value']
+                            if value is not None and len(value)>0:
```
acs: Just `if value` must cover both cases

grimoire_elk/elk/jira.py
jgbarah: Ooops. Yes. Sorry....

grimoire_elk/elk/jira.py
jgbarah: I had it that way for readability, but I will change it.

jgbarah:
Thanks for the comments. I think they should now be addressed.

PR #122, opened 2017-11-20
Add support for Elastic 6

acs:
This PR fix kidash so it works now with Elasticsearch 6 also (also supporting Elasticsearch 2.x and 5.x).

During this support I have refactored the code to simplify it and to make it easier to add the support.

Also, a new option "--data-sources" is now available to load dashboards with all data sources like Overview and Data Status, and removing all widgets except those related to the data sources included as params.

For example:

`./kidash.py -g  --import ~/devel/panels/json/data_status.json --data-sources github git
`

### DIFF
--- a/grimoire_elk/panels.py
+++ b/grimoire_elk/panels.py
@@ -29,7 +29,6 @@ import logging
 import os
 import os.path
 import pkgutil
-import sys
 
 from .elk.elastic import ElasticSearch
 from .elk.utils import grimoire_con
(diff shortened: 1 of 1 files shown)


alpgarcia:
Thanks @acs. I would separate those commits by functionality:
* All commits related to ES6 could remain together as they are intended to upgrade the tool to ES6.
* Kidash commit should go in a separate PR.

Besides, I'd say it would be very helpful to have a bit more context in commit messages. For instance, `Fix library to work with Elasticsearch6` is a great subject, but I miss a brief explanation on why it was needed. Basically I would add a brief text to all commits just saying something like: `This change fixes/add this particular problem`. It would make clear what the change does without the need of looking at the diff. 

Another example could be:
```
[panels] Support the export of Kibana6 panels

* URLs in ES6 have changed so import code needed to be updated keeping
  backwards compatibility.
* Item types have been removed in ES. Added code to remove types when
  dealing with ES6 panels. 
...
```

Thanks!

acs:
@alpgarcia good point about splitting the PR in one for ES6 (panels) and the other one for kidash. I will create two new PRs and close this one once this is done!

And during this process, I will improve the commits messages following your suggestions. Thank you!

acs:
@alpgarcia commits messages reworded. This PR has only now the ES6 stuff. The new PR with widgets cleaning is https://github.com/grimoirelab/GrimoireELK/pull/124 and it needs that this PR is first accepted.

alpgarcia:
LGTM! Thanks!

PR #126, opened 2017-11-24
Remove crates empy fake repository needed by mordred

acs:
The perceval backend for crates does not need any repository param.
    But we need to add an empty repository to projects.json so the backend
    is executed by mordred. Before creating the real perceval backend we need
    to remove this fake repository param:
    
    Contents of projects.json:
    
    {
        "grimoirelab": {
            "crates": [""]
    ...


alpgarcia:
LGTM!

acs:
Thanks @alpgarcia

PR #127, opened 2017-11-24
[index_mapping] Add an option so the elasticsearch for read and write could be different

acs:
This use case is needed to test if there are problems with the items
write when are are reading the items from a Elasticsearch that the logs
can not be analyzed (like a customer AWS ES).


acs:
This PR is included also in https://github.com/grimoirelab/GrimoireELK/pull/128. Closing it

PR #128, opened 2017-11-24
Fix raw mappings for bugzillarest and nntp

acs:
As all strings in raw indexes are stored as keywords there are some
issues with large strings that can not be stored as keywords. In
this case the string field must not be indexed.


acs:
This PR has been broken down in:

https://github.com/grimoirelab/GrimoireELK/pull/129
https://github.com/grimoirelab/GrimoireELK/pull/130
https://github.com/grimoirelab/GrimoireELK/pull/131

Closing this one!

PR #129, opened 2017-11-26
Index mapping ++

acs:
Improvements to index_mapping tool:

* Read and write Elasticsearches could be in different hosts
* Can use search-after for scrolling the index (alpha)
* Use sync bulk upload to check that all items are written


zhquan:
LGTM

PR #131, opened 2017-11-26
Bulk sync packs improvements

acs:
During the bulk upload in sync mode, items are now packed so after each pack, the number of items really indexed as new can be check in order to detect duplicate items or problems adding the items (for example for mappings issues).

### DIFF
--- a/grimoire_elk/elk/elastic.py
+++ b/grimoire_elk/elk/elastic.py
@@ -135,37 +135,59 @@ class ElasticSearch(object):
         return new_items
 
     def bulk_upload_sync(self, items, field_id, sync=True):
-        ''' Upload in controlled packs items to ES using bulk API
+        ''' Upload items in packs to ES using bulk API
             and wait until the items appears in searches '''
 
         # After a bulk upload the searches are not refreshed real time
         # This method waits until the upload is visible in searches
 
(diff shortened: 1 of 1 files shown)


grimoire_elk/elk/elastic.py
```diff
@@ -135,37 +135,59 @@ def bulk_upload(self, items, field_id):
         return new_items
 
     def bulk_upload_sync(self, items, field_id, sync=True):
-        ''' Upload in controlled packs items to ES using bulk API
+        ''' Upload items in packs to ES using bulk API
```
albertinisg: Standardize the quotes

grimoire_elk/elk/elastic.py
```diff
@@ -134,35 +135,58 @@ def bulk_upload(self, items, field_id):
...
-                break
+        def current_items():
+            """ Current number of items in the index """
+            res = self.requests.get(self.index_url+'/_search?size=1')
+            res.raise_for_status()
+            return res.json()['hits']['total']
+
+        def wait_index(total_expected):
+            """ Wait until ES has indexed all items """
+            # Wait until in searches all items are returned
```
albertinisg: I guess this should be part of the function description

PR #133, opened 2017-11-27
[enrich] Check if an identity exists before trying to get her profile

acs:


alpgarcia:
LGTM!

Thanks so much!

PR #140, opened 2017-11-29
[tests][crates] Add testing for crates data source

acs:


acs:
I am self reviewing this PR because it is trivial the change and we need the test env for tetsing the pep8 refactoring!

PR #147, opened 2017-12-07
[enrich][sortinghat] Add identity domain using the profile information

acs:
There was a case in which the domain was not added to the identities
field during enrichment. This fix covers it.


alpgarcia:
LGTM, thanks!

PR #148, opened 2017-12-12
Arthur integration branch

acs:
This is a branch which includes the integration work with arthur that has been done during the last months. It is not ready so it won't be merged during the next days ... but all the integration work will be finished here!


acs:
This is the PR that adds the arthur support to GrimoireELK.

PR #149, opened 2017-12-19
[autorefresh] Limit the max number of identities to be searched for refreshing

acs:
If the number of identities is higher than elastic.max_items_clause (1000)
then all the index is refreshed. The max number of clauses that Elasticsearch
support by default is 1024, so with this limit we won't have issues
executing the search query with all the identities to find their related
items.

### DIFF
--- a/grimoire_elk/arthur.py
+++ b/grimoire_elk/arthur.py
@@ -246,20 +246,40 @@ def refresh_projects(enrich_backend):
 
 
 def refresh_identities(enrich_backend, filter_author=None):
+
+    def update_items(new_filter_author):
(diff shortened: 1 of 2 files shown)


acs:
Refreshing items for more the 1k identities could also be several hours. We need to measure them. It is easy in any case to group the identities in 1k groups and do the refreshing for them. And iterate. Let's talk about it tomorrow. Thanks @albertinisg !

acs:
The new PR group the identities in 1000 items list and refresh the index for this group of identities. With this approach there is no limit in the number of identities to be refreshed. @albertinisg dose it solve your concerns?

PR #151, opened 2017-12-21
Add Git Areas of Code study

alpgarcia:
Areas of code study is based on Ceres library, that should be
in PYTHONPATH when running study. A symbolic link is included
in utils directory, pointing to repository root, where ceres
repo should be included (or a symbolic link to it).

In this very first version you will also need a file named `.aoc_settings` with:
```
[ElasticSearch]
user=
password=
host=
port=
path=
index_git_raw=

user_output=
password_output=
host_output=localhost
port_output=
path_output=
index_git_output=


[SortingHat]
db_user=
password=
db_name=
host=
port=

[General]
projects=
size=
# inc: 'True' ignoring case. Other strings are considered 'False'.
inc=True
```


acs:
I have tested it and it works once all the configuration and dependencies are satisfied. Great!

sduenas (changes requested):
Update the PR to use the latest `ceres` package. Don't include it in `gelk` as symbolic reference.

alpgarcia:
Closing it. Following this integration at #228

PR #154, opened 2017-12-28
[setup] Add kingarthur as a dependency

jgbarah:
Removed all extra dependency on arthur, and updated some other dependencies.


jgbarah:
I had not realized the problems with flake8. I think they're fixed now.

jgbarah:
Mmm. Still some tests fail, but I think they are not related to my change...

acs:
Sometimes the connection to Travis Elasticsearch service fails ... executing another time the test to check this is the current case!

PR #157, opened 2018-01-15
Support for Elasticsearch 6, from elastic6 branch in acs/GrimoireELK

jgbarah:
Resolved merging.
Added a new function in ElasticSearch class to detect the release of the Elasticsearch instance.
Added testing, for that new class.
Added check when uploading default mapping, depending on the version of Elasticsearch.
Changed "string" to "keyword" or "text" (depending if analyzed or not analyzed) in mappings for backends that needed it.
Added some debug messages for better tracking when uploading and fetching items.

Please, run all tests, I couldn't do it for all datasources.


acs:
First, we need to fix the error with the tests. It seems that httpretty is no installed in travis. We need to do it. I would change the name "test_elk_elsticsearch.py"  to "test_elk_elasticsearch.py". Could you add this dependency to `.travis.yml`?

Also, we need to be sure that gelk continues working with ES 2.x because GrimoireELK in production will publish raw indexes in an env with ES 2.x. 

From a global point of view, I need to check that we have not lost anything in the merge, because there were some large conflicts to solve.

But I will try to look in more deep ASAP.

jgbarah:
> First, we need to fix the error with the tests. It seems that httpretty is no installed in travis. We need to do it. I would change the name "test_elk_elsticsearch.py" to "test_elk_elasticsearch.py". Could you add this dependency to .travis.yml?

Dependency added to .travis.yml

Name changed from "test_elk_elsticsearch.py" to "test_elk_elasticsearch.py" ;-)

> Also, we need to be sure that gelk continues working with ES 2.x because GrimoireELK in production will publish raw indexes in an env with ES 2.x.

Done. I hope it has now even better support for enriched indexes with ES 2.x

> From a global point of view, I need to check that we have not lost anything in the merge, because there were some large conflicts to solve.

Yes, test it as much as possible...

jgbarah:
OK, I've just realized some tests are failing. I think that's because the change in mappings. Hold on for a minute, I will try to fix this during today.

jgbarah:
When I run the tests, I still see that those on the enriched index for Twitter fail. The problem seems to be that the enriched index is queried (for example, when updating identities) using the `@timestamp` field, which does not exist in the enriched index. I've tried to find where it is set, and change it to `timestamp`, but in that case tests on the raw index fail (apparently because they need `@timestamp` in the query, and they are using 'timestamp'). This is weird, because apparently the raw index should not be affected... The lines I refer to are (in file `elastic_items.py`):

```python
            # Order the raw items from the old ones to the new so if the
            # enrich process fails, it could be resume incrementally
            order_query = ''
            order_field = None
            if self.perceval_backend:
                order_field = self.get_incremental_date()
            elif self.get_connector_name() == 'twitter':
                order_field = '@timestamp'
            if order_field is not None:
                order_query = ', "sort": { "%s": { "order": "asc" }} ' % order_field
```

Any idea?

jgbarah:
New update. Now, all tests passed in my laptop using Elasticsearch 6.0, 5.0, and 2.4.

Ufff

;-)

jgbarah:
Hmmm, I cannot reproduce the errors shown in the latest Travis run. Maybe something wrong in Travis? Checks keep passing in my laptop with all ES versions...

acs:
Sometimes travis fails to start correctly the elasticsearch service ... let's retry the tets!

https://travis-ci.org/grimoirelab/GrimoireELK/builds/330493737?utm_source=github_status&utm_medium=notification

jgbarah:
It seems only flake8 is failing, on some kidash files. But those files are removed in #158, so please accept this patch. flake8 will be happy when you accept it too.

grimoire_elk/elk/enrich.py
```diff
@@ -452,12 +419,11 @@ def get_last_offset_from_es(self, _filters=[]):
...
-        """ Mappings for enriched indexes """
-
-        mapping = '{}'
-
-        return {"items": mapping}
+#    def get_elastic_mappings(self):
+#        """ Mappings for enriched indexes """
+#
+#        mapping = '{}'
+#        return {"items": mapping}
```
acs: Remove this comments if this code is not used anymore

PR #158, opened 2018-01-18
[kidash] Remove kidash (moved to a new repo as chaoss/grimoirelab-kidash)

jgbarah:
This should be the final step for removing kidash from this repo.


jgbarah:
I've rebased the change. I hope it works now directly on top of master.

PR #159, opened 2018-01-20
[filters] Fix bug checking filters contents during incremental enrich…

acs:
…ment


acs:
@jgbarah I have already included this fix because it breaks the incremental enrichment. I add you as reviewer so you are aware of it!

PR #168, opened 2018-01-30
[elastic] Track storage exceptions when uploading data to ES

valeriocos:
This patch enables the ElasticSearch component to handle storage exceptions, which were ignored before.


acs:
@valeriocos could you check the errors with the tests and fix them? Thanks!

PR #172, opened 2018-02-01
[enrich][git][demography] Fix the demography study logic so it works in incremental mode always

acs:
The key is that the new enriched commits are the ones that don't have
the demography study fields: author_min_date and author_max_date. So only
the authors of these new commits must be refreshed.


acs:
To include it in elasticgirl.30.1 hotfix release I am auto merging it. Please @jgbarah take a look to it in any case.

PR #173, opened 2018-02-06
[git] Fix for refresh identities in pair programming git

acs:
Fix bug when refreshing identities in pair programming git.

The uuid in this case is not unique (a raw commit could create several enriched commits with the same uuid) so a new git_uuid has been added which is unique.


sanacl:
LGTM.

PR #175, opened 2018-02-08
[jenkins] Remove data.changeSet.items.comment from index

valeriocos:
This fix prevents to index the comment on a changeSet. Due to some large comments on changeSets, some Jenkins items caused ElasticSearch to fail in storing them.


acs (approved):
LGTM

PR #176, opened 2018-02-08
[enrich] Track storage exceptions when uploading enriched data to ES

valeriocos:
This patch enables the ElasticSearch component to handle storage exceptions when uploading enriched data.


acs (approved):
LGTM

PR #178, opened 2018-02-08
[enrich][crates] Return always a correct SH identity

acs:


grimoire_elk/elk/crates.py
```diff
@@ -87,6 +87,10 @@ def get_identities(self, item):
     def get_sh_identity(self, item, identity_field=None):
         identity = {}
 
+        identity['username'] = None
+        identity['email'] = None
```
valeriocos: Is identity['email'] needed?

acs: Yes, all the SH fields must exist always: username, email, name.

acs:
Merging this one ... is s minor change moving code.

PR #179, opened 2018-02-09
[ocean] Add warning when json items are not inserted to ES

valeriocos:
This patch adds warnings to the method _items_to_es within ocean/elastic.py. It is thrown when a mismatch exists between the number of json items in input and the number of the inserted ones in the ES storage

### DIFF
--- a/grimoire_elk/ocean/elastic.py
+++ b/grimoire_elk/ocean/elastic.py
@@ -204,7 +204,6 @@ class ElasticOcean(ElasticItems):
         self.feed_items(items)
 
     def feed_items(self, items):
-
         task_init = datetime.now()
 
         items_pack = []  # to feed item in packs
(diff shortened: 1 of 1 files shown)


grimoire_elk/ocean/elastic.py
```diff
@@ -204,7 +204,7 @@ def feed(self, from_date=None, from_offset=None, category=None, latest_items=Non
         self.feed_items(items)
 
     def feed_items(self, items):
-
+        
         task_init = datetime.now()
```
acs: You should remove this tab

grimoire_elk/ocean/elastic.py
```diff
@@ -248,4 +248,9 @@ def _items_to_es(self, json_items):
         field_id = self.get_field_unique_id()
 
         inserted = self.elastic.bulk_upload_sync(json_items, field_id)
+
+        if len(json_items) != inserted:
+            missing = len(json_items) - inserted
+            logger.warning("%s JSON items not inserted to ES", str(missing))
+
```
acs: I will add to the log the name of the method in which it has occur. If not it is a bit generic.

acs (approved):
LGTM

PR #180, opened 2018-02-09
[tests] Update tests data

valeriocos:
This patch provides a new dataset of tests, extracted from the latest implementations of Perceval backends. Previous tests are temporary moved to the folder tests/old_data


valeriocos:
Travis fails because this PR requires changes contained in the PRs 181-192

PR #181, opened 2018-02-09
[ocean] Add GitLab backend

valeriocos:
This code includes the ocean backend for GitLab


grimoire_elk/ocean/gitlab.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# StackExchange Ocean feeder
```
acs: This comment is wrong

acs (changes requested):

PR #182, opened 2018-02-09
[ocean] Add PuppetForge backend

valeriocos:
This code includes the ocean backend for PuppetForge


grimoire_elk/ocean/puppetforge.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# StackExchange Ocean feeder
```
acs: this comment is wrong

acs (changes requested):

PR #190, opened 2018-02-09
[elk] Update enrich_items method of Mediawiki for new tests

valeriocos:
The enrich_items method should return the number of items processed. This value is then checked against the number of items in the raw index in the test.


acs (approved):
LGTM

PR #191, opened 2018-02-09
[elk] Handle NNTP uppercase Date field

valeriocos:
This patch allows to handle through the MBox backend the case when the Date attribute of the JSON items appears in uppercase


acs (approved):
LGTM.

PR #193, opened 2018-02-11
Refresh identities git p2p: add a new param pair_programming

acs:


acs:
Auto review to sync carefully with the PR that uses it in mordred.

PR #195, opened 2018-02-12
[ocean] Fix mappings to index Askbot data

valeriocos:
This patch prevents to index Askbot answer summaries, and the content of answer comments and their summaries, which may cause failures when uploading the data to the ES storage


grimoire_elk/ocean/askbot.py
```diff
@@ -48,11 +44,24 @@ def get_elastic_mappings(es_major):
...
+                                            }
+                                        }
+                                    },
+                                    "summary": {
+                                        "type": "text",
+                                        "index": false
+                                    }
+                                }
                             },
-                            "author": {
```
acs: Why have you removed this mapping @valeriocos ? Not sure if we need it ... I would not change it unless we are totally sure we don't need it.

acs (approved):
LGTM

PR #196, opened 2018-02-12
[ocean] Fix mappings to index Bugzilla data

valeriocos:
This patch prevents to index long and short descriptions of Bugzilla issues, which  may cause failures when uploading the data to the ES storage


acs (approved):
LGTM

PR #197, opened 2018-02-12
[ocean] Fix mappings to index BugzillaRest data

valeriocos:
This patch prevents to index summary issues, the content of comments and their descriptions and summaries of attachments. These fields may contain large text that causes ES to fail during data uploading.


acs (approved):
LGTM

PR #198, opened 2018-02-12
[ocean] Enhance mappings to index Crates data

valeriocos:
This patch prevents to index fine-grained versions and downloads information, which may cause ES to fail in handling to many indexes


acs (approved):
LGTM

PR #199, opened 2018-02-12
[discourse] Fix mappings to index Discourse data

valeriocos:
This patch prevents to index text content within Discourse data, which may cause failures when uploading the data to ES


acs (approved):
LGTM

PR #200, opened 2018-02-12
[ocean] Fix mappings to index DockerHub data

valeriocos:
This patch prevents to index the descriptions of docker images, which may cause failures when uploading the data to ES


grimoire_elk/elk/dockerhub.py
```diff
@@ -43,32 +43,23 @@ def get_elastic_mappings(es_major):
...
-                      "index": "analyzed"
-                    }
-               }
-            } """
+        mapping = """
+        {
+            "properties": {
+                "description": {
+                    "type": "text",
+                    "index": false
```
acs: We want to index this field. And the same for the two below.

acs (changes requested):

grimoire_elk/elk/dockerhub.py
valeriocos: a description may be too long and trigger a **Lucene’s term byte-length limit of 32766** exception. Why do we need to index this field?
We could improve this mapping using the option [ignore-above](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html)

acs (approved):
LGTM

PR #201, opened 2018-02-12
[ocean] Fix mappings to index Gerrit data

valeriocos:
This patch prevents to index possible large text fields returned by the Gerrit API, which may cause failures when uploading the data to ES


acs:
Hmmm, I won't work in possible causes of failures. But let's talk about it! Thanks!

grimoire_elk/elk/gerrit.py
```diff
@@ -45,33 +45,23 @@ def get_elastic_mappings(es_major):
...
-                   "timeopen": {
-                      "type": "double"
-                   }
-                }
+        mapping = """
+        {
+            "properties": {
+               "summary_analyzed": {
+                  "type": "text",
+                  "index": false
```
acs: We want to index this field.

acs (changes requested):

grimoire_elk/elk/gerrit.py
valeriocos: Long summaries may trigger a Lucene exception (Lucene’s term byte-length limit of 32766). Why we have to index this field?

acs (approved):
LGTM

PR #202, opened 2018-02-12
[ocean] Remove ES 2.x mappings to index Git data

valeriocos:
This patch removes mappings for ES 2.x

### DIFF
--- a/grimoire_elk/elk/git.py
+++ b/grimoire_elk/elk/git.py
@@ -65,8 +65,8 @@ class Mapping(BaseMapping):
             {
                 "properties": {
                    "message_analyzed": {
-                      "type": "text",
-                      "fielddata": true
+                        "type": "text",
+                        "index": true
                    }
                }
             }"""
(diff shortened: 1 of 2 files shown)


acs:
Why are you removing the mappings for ES 2.x? We must support this version yet.

valeriocos:
Most Ocean backends don't provide this feature. I thought that those mappings were dead code

acs:
This removal can cause regressions so I won't do that.

grimoire_elk/ocean/git.py
```diff
@@ -39,38 +39,21 @@ def get_elastic_mappings(es_major):
         :returns:        dictionary with a key, 'items', with the mapping
         """
 
-        if es_major == '2':
```
acs: This removal can cause regressions so I won't remove the 2.x support.

grimoire_elk/elk/git.py
```diff
@@ -75,8 +75,8 @@ def get_elastic_mappings(es_major):
             {
                 "properties": {
                    "message_analyzed": {
-                      "type": "string",
-                      "index": "analyzed"
+                        "type": "string",
+                        "index": "analyzed"
```
acs: Some strange spaces here ... could you review it?

acs (commented):
LGTM

acs (approved):
LGTM

PR #203, opened 2018-02-12
[ocean] Replace Jenkins mapping property for changeSet.items.comment

valeriocos:
This patch changes how the comments on changeSet items are prevented to be indexed. Thus, the old value "not_analyzed" of the mapping "index" is replaced by "false"

### DIFF
--- a/grimoire_elk/ocean/jenkins.py
+++ b/grimoire_elk/ocean/jenkins.py
@@ -41,27 +41,29 @@ class Mapping(BaseMapping):
         :returns:        dictionary with a key, 'items', with the mapping
         """
 
-        mapping = '''
-         {
-            "dynamic":true,
-                "properties": {
-                    "data": {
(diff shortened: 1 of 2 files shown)


grimoire_elk/ocean/jenkins.py
```diff
@@ -61,7 +61,7 @@ def get_elastic_mappings(es_major):
                                         "properties": {
                                             "comment": {
                                                 "type": "text",
-                                                "index": "not_analyzed"
+                                                "index": false
                                             }
```
acs: Not sure here ... I think we want to index this field.

valeriocos: Long comments can trigger a Lucene exception (Lucene’s term byte-length limit of 32766). Why we have to index this field?

acs (approved):
LGTM

PR #204, opened 2018-02-12
[ocean] Fix mappings to index Mediawiki data

valeriocos:
This patch prevents to index the content of revision comments, which may cause failures when uploading the data to ES


grimoire_elk/elk/mediawiki.py
```diff
@@ -44,25 +44,19 @@ def get_elastic_mappings(es_major):
         :returns:        dictionary with a key, 'items', with the mapping
         """
 
-        if es_major != '2':
```
acs: I won't remove the support for 2.x because it can cause regressions.

acs (approved):
LGTM

PR #205, opened 2018-02-13
[ocean] Fix mappings to index Meetup data

valeriocos:
This patch prevents to index the content of comments and author bios, which may cause failures when uploading the data to ES


grimoire_elk/elk/meetup.py
```diff
@@ -43,38 +43,25 @@ def get_elastic_mappings(es_major):
         :param es_major: major version of Elasticsearch, as string
         :returns:        dictionary with a key, 'items', with the mapping
         """
-
-        if es_major != '2':
-            mapping = """
```
acs: I won't remove the ES 2.x support because it can cause regressions.

acs (changes requested):

acs (commented):
Hmmm, I can not see ES 2.x support readded.

valeriocos:
I have just updated the PR

acs (approved):
LGTM

PR #206, opened 2018-02-13
[ocean] Fix mappings to index MozillaClub data

valeriocos:
This patch prevents to index the event descriptions, which may cause failures when uploading the data to ES


grimoire_elk/elk/mozillaclub.py
```diff
@@ -42,52 +42,30 @@ def get_elastic_mappings(es_major):
         :returns:        dictionary with a key, 'items', with the mapping
         """
 
-        if es_major != '2':
```
acs: I won't remove the ES 2.x support because it can cause regressions.

acs (changes requested):

acs (approved):
LGTM

PR #207, opened 2018-02-13
[ocean] Fix mappings to index NNTP data

valeriocos:
This patch prevents to index the subject and the body of NNTP messages which may cause failures when uploading the data to ES


grimoire_elk/elk/nntp.py
```diff
@@ -22,11 +22,44 @@
...
+        mapping = """
+        {
+            "properties": {
+                "Subject": {
+                    "type": "text",
+                    "index": false
+                },
+                "Subject_analyzed": {
+                    "type": "text",
+                    "index": false
```
acs: we want to index this field

acs (changes requested):

grimoire_elk/elk/nntp.py
valeriocos: Long subjects may trigger a Lucene exception (Lucene’s term byte-length limit of 32766). Why we have to index this field?

acs (approved):
LGTM

PR #208, opened 2018-02-13
[ocean] Modify Pipermail input data for test

valeriocos:
This patch increases the size of a body message of an email within the test data to ensure that the defined mapping works properly


acs (approved):
LGTM

PR #210, opened 2018-02-13
[ocean] Fix mappings to index Remo data

valeriocos:
This patch prevents to index description of Remo events, which may cause failures when storing the data to ES


acs (approved):
LGTM

PR #211, opened 2018-02-13
[ocean] Fix mappings to index RSS data

valeriocos:
This patch prevents to index summary and content information, which may cause failures when uploading the data to ES


grimoire_elk/elk/rss.py
```diff
@@ -43,25 +43,19 @@ def get_elastic_mappings(es_major):
         :returns:        dictionary with a key, 'items', with the mapping
         """
 
-        if es_major != '2':
-            mapping = """
```
acs: I won't remove the ES 2.x support because it can cause regressions.

acs (changes requested):

acs (approved):
LGTM

PR #212, opened 2018-02-13
[ocean] Fix mappings to index Stackexchange data

valeriocos:
This patch prevents to index the content of questions and answers which may cause failures when storing data to ES


acs (approved):
LGTM

PR #213, opened 2018-02-13
[travis] Add delay to make ES start

valeriocos:
This code introduces 15 second delay before the build script starts to let ElasticSearch starts. Thus it avoids the build to stall.


acs (approved):
LGTM

PR #214, opened 2018-02-14
[elk] Remove gmane connector

valeriocos:
This code removes gmane from the connectors, since Perceval and Arthur don't provide support for it anymore


acs (approved):
LGTM

PR #215, opened 2018-02-14
[elk] Fix enrichment mappings for Slack

valeriocos:
This patch fixes the enrichment mappings for Slack items, which were causing ES to fail when inserting boolean values to a numeric field.


acs (changes requested):
`user_data_bot` must be a boolean value because this is coming directly from SortingHat profile info. we need to respect the SH profile types to have the same approach for all SH fields in all places in gelk.

valeriocos:
ok, I see. I close this PR. To solve the issue we have just to recreate the enriched index for the data source that is failing

PR #216, opened 2018-02-14
[tests] Add tests for Askbot

valeriocos:
This code isolates the code to test loading of raw, enriched and identities information


acs:
@valeriocos and me have decided to analyze the approach of passing then backend to be tested as a param. So we won't merge this PR yet.

PR #220, opened 2018-02-23
[tests] Refactor tests for backends/connectors

valeriocos:
This code refactors the code used to tests oceans and elk connectors. Now tests can be executed for a subsets of connectors, selected via the config file. Furthermore, a new test checks that the connectors available in elk and ocean packages are aligned with the ones listed in the grimoire_elk/utils.py file.


acs (approved):
LGTM

PR #223, opened 2018-02-26
[rss] Reduce test data for RSS

valeriocos:
This code reduces the size of the test data for the RSS connector. The current size may affect the time required by Travis to check the build, causing in some cases to not produce any output.


acs (approved):
LGTM

PR #226, opened 2018-02-27
[ocean] Fix mappings to index Twitter data

valeriocos:
This patch reduces the number of indexes created when inserting Twitter data, thus preventing hitting the limit of total fields in index. For each Twitter items, the fields below are not indexed:
- user.entities
- hashtags.indices
- metadata
- entities.symbols
- entities.urls
- entities.user_mentions


acs (approved):
LGTM

Issue #224, opened 2018-02-26
Add Archive mode


sduenas:
Perceval 0.9.11 adds `Archive` mode, removing the `Cache` one. Grimoire ELK still uses the `Cache`, so newer versions of Perceval cannot be used.

The purpose of this issue is to integrate `Archive` in GELK and remove the Cache. `Cache` never worked as expected so it's safe to remove it.

sduenas:
Fixed by #227

Issue #222, opened 2018-02-25
Latest perceval seems to break grimoire_elk


jgbarah:
When trying to use the master/HEAD for perceval, perceval-mozilla and perceval-opnfv, it seems grimoire_elk fails. For example, when running p2o.py:

```
2018-02-23 17:08:00,421 - grimoire_elk.arthur - ERROR - Error feeding ocean 'GitCommand' object has no attribute 'backend'
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/arthur.py", line 121, in feed_backend
    backend = backend_cmd.backend
AttributeError: 'GitCommand' object has no attribute 'backend'
```

I can patch grimoire_elk/arthur.py to get rid of this error:

```
a/grimoire_elk/arthur.py
+++ b/grimoire_elk/arthur.py
@@ -118,7 +118,7 @@ def feed_backend(url, clean, fetch_cache, backend_name, backend_params,
     try:
         backend_cmd = klass(*backend_params)
-        backend = backend_cmd.backend
+        backend = backend_cmd.BACKEND
         ocean_backend = connector[1](backend, fetch_cache=fetch_cache, project=project)
         logger.info("Feeding Ocean from %s (%s)", backend_name, backend.origin)
```

But I still get more errors:

```
p2o.py --enrich --index git_raw --index-enrich git -e https://admin:admin@localhost:9200 --no_inc --debug git https://github.com/grimoirelab/perceval.git
....
2018-02-23 23:27:45,554 Error feeding ocean from git (<property object at 0x7f188b7b6778>): fetch() missing 1 required positional argument: 'self'
Traceback (most recent call last):
  File "/home/jgb/src/jgbarah/mordred/grimoire_elk/arthur.py", line 197, in feed_backend
    ocean_backend.feed()
  File "/home/jgb/src/jgbarah/mordred/grimoire_elk/ocean/elastic.py", line 202, in feed
    items = self.perceval_backend.fetch()
TypeError: fetch() missing 1 required positional argument: 'self'
Traceback (most recent call last):
  File "/tmp/gl/bin/p2o.py", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/home/jgb/src/jgbarah/GrimoireELK/utils/p2o.py", line 73, in <module>
    args.arthur)
  File "/home/jgb/src/jgbarah/mordred/grimoire_elk/arthur.py", line 219, in feed_backend
[...]

PR #227, opened 2018-02-28
[archive] Allow handling archive

valeriocos:
This PR targets issues #224 and #222. It enables the use of the archive in GElk. Note that (currently) only storing data to archive is supported (it does not allow to fetch data from archives).


acs (approved):
LGTM

PR #228, opened 2018-02-28
[elk] Add areas of code study

alpgarcia:
Some new classes added in `ceres_base.py` to ease integration of
cereslib based enrichers.

`sortinghat.py` gelk module was refactored to `sortinghat_gelk.py`
due to importing problems when using `cereslib`, because cereslib
uses `sortinhat` from gelk and, at the same time, gelk was using
`sortinghat` from sortinhat project. This needs a new version of
 cereslib package according to: 
https://github.com/dicortazar/ceres/pull/13


grimoire_elk/elk/mappings/git_aoc.json
```diff
@@ -0,0 +1,113 @@
...
+        "removedlines": {
+          "type": "long"
+        },
+        "repository": {
+          "type": "keyword"
+        }
+      }
+    }
+  }
+}
```
acs: Why are we storing the mapping in this directory? We are not following this approach in any other case.

alpgarcia: I didn't know where to store it and I don't want to put the explicit mapping in the source file for readability. I just tried to have it closer to the file using it.

The thing is that all enrichers coming from `cereslib` have their own explicit mappings.

grimoire_elk/elk/ceres_base.py
```diff
@@ -0,0 +1,328 @@
...
+        :param hits_block:
+        :return: hits blocks as they come, updating their metadata__enriched_on field. Namedtuple containing:
+            - processed: number of processed hits
+            - out_items: a list containing items ready to be written.
+        """
+
+        out_items = []
+
+        for hit in hits_block:
+            hit['_source']['metadata__enriched_on'] = dt.utcnow().isoformat()
```
sduenas: I would use, `grimoirelab-toolkit` functions to work with dates, but that's only my opinion.

grimoire_elk/elk/study_ceres_aoc.py
```diff
@@ -0,0 +1,176 @@
+#!/usr/bin/python
```
sduenas: Change this to use `env`

sduenas (changes requested):

grimoire_elk/elk/ceres_base.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def __init__(self, in_connector, out_connector, block_size):
+        """
+
+        :param in_connector: connector for reading source items from
+        :param out_connector: connector to write processed items to
+        :param block_size: number of items to retrieve for each processing block
+        """
+
+        self.__in = in_connector
```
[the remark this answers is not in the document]
alpgarcia: You are right, they should be protected.

grimoire_elk/elk/ceres_base.py
```diff
@@ -0,0 +1,328 @@
...
+    processing them and writing the resulting processed data. As this is a very general
+    procedure, it can be also extended to create other non-cereslib based processors.
+    """
+
+    ProcessResults = namedtuple('ProcessResults', ['processed', 'out_items'])
+
+    def __init__(self, in_connector, out_connector, block_size):
+        """
+
+        :param in_connector: connector for reading source items from
```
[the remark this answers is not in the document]
alpgarcia: Do you mean class fields?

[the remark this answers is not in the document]
sduenas: Init parameters.

alpgarcia:
I've addressed all your comments @sduenas. Once you've reviewed them I'll reorganize these commits to have:
- One commit for ceres base.
- One commit for areas of code (refactoring could be include here or in a separated commit).
- One commit for travis setup.
- One commit for updating requirements.

grimoire_elk/elk/mappings/git_aoc.json
acs: It is ok for me then! Let's go ahead with this approach. But at some point we must unify criteria about mappings in gelk and cereslib.

sduenas (approved):
LGTM

grimoire_elk/elk/git.py
```diff
@@ -697,3 +703,19 @@ def enrich_demography(self, enrich_backend, no_incremental=False):
         self.elastic.bulk_upload(author_items, "_item_id")
 
         logger.debug("Completed demography enrich from %s" % (self.elastic.index_url))
+
+    def enrich_areas_of_code(self, enrich_backend, no_incremental=False):
+        logger.info("[Areas of Code] Starting study")
+
+        # Creating connections
+        es = Elasticsearch([self.elastic.url])
+        in_conn = ESPandasConnector(es_conn=es, es_index="git_areas_of_code-raw", sort_on='metadata__timestamp')
```
acs: Here it is using as Elasticsearch for reading the data from, the same one than the Elasticsearch used for the enriched indexes. In general, these Elasticsearch are different. Is it possible to access the url for the ES with the raw index? Let me check ...  this is how the items are enriched from the raw index. The data for the raw index is passed as a param, so the Enrich object does not have information about where the raw index Elasticsearch is.

```
    def enrich_items(self, ocean_backend, events=False):
        max_items = self.elastic.max_items_bulk
        current = 0
        total = 0
        bulk_json = ""

        items = ocean_backend.fetch()
```

So I will leave as a requisite for this study that the ES is the same for raw and enriched indexes, and if in the future we try to use it in other cases, we need to add support so the Enrich class can obtain in some way where the raw data is.

acs (approved):

grimoire_elk/elk/git.py
alpgarcia: Maybe we could open an enhancement issue about `study configuration` and discuss it there as future work.

acs: Yes! My guess is that this config must be passed to the study as params, yes! Good point. Opening a new issue then:

https://github.com/chaoss/grimoirelab-elk/issues/232

PR #229, opened 2018-03-02
[enrich][git] Fix incremental demography update for already existing authors

acs:
Readded the logic to find refresh authors commits which new commits since the
last execution of the demography in order to update the author_max_date for them.


valeriocos (approved):
LGTM

PR #230, opened 2018-03-02
[archive] Add support to fetch data from archives

valeriocos:
This code allows to fetch data from a **single** archive


acs (approved):
LGTM

PR #233, opened 2018-03-06
[elastic] Refactor how items are stored in ES

valeriocos:
This PR enables to store enriched items using the common method safe_put_bulk. Thus, controlling possible failures and returning the number of inserted items


acs (approved):
LGTM

PR #234, opened 2018-03-06
[tests] Split tests for backends

valeriocos:
This PR splits the tests for backends to single files. Thus, it allows to define ad-hoc tests for every backend


acs (approved):
LGTM

PR #237, opened 2018-03-07
[elastic] Delete bulk_upload_sync method

valeriocos:
This code removes the bulk_upload_sync method and replaces its calls with the ones of bulk_upload method, which now takes care of controlling that the changes made are visible to search (by adding the param ?refresh=true to the request).


acs (approved):
We are using now "?refresh=true" in the normal bulk_upload, which is the same feature implemented in bulk_upload_sync, so we can remove it. I am not sure if we should use `wait_for` instead `true` but the current approach is worse that both of them, so let's change it!

PR #238, opened 2018-03-08
[git] Add a new time_to_commit field to enriched index.

### LINKED ISSUE #235: Add tIme-to-commit field in git indexes
In chaoss/grimoirelab#79 @BenLloydPearson says:

> One thing I have used a lot that doesn't seem to be supported in GrimoireLab is a field that provides the amount of time it takes for code to be committed upstream. This is the difference between the author_date and commit_date fields. In the past, I've done this with a scripted field in Kibana, but from my understanding this is very inefficient, and it also tends to disappear frequently after running Mordred (it seems to do something that wipes out scripted fields when I run it).
>
> Kibana offers a basic time difference calculation, but it outputs a number of milliseconds and needs some formatting to be made human readable. I've typically expressed this in a number of days to the nearest tenth (e.g. 14.2 days), but hours might also work.
>
> Would it be possible to add a field for something like time_to_commit?

I think this is certainly doable and convenient. The value of both author and commit time depend a bit on how the commit is merged, but anyway, this can be useful.

WRT the unit of time, I would make it be seconds, and then use representation to convert it to other units, such as days. But we can discuss that in this ticket.

BTW, pull requests for this are welcome! It shouldn't be difficult to implement this.

jgbarah:
Time to commit can be computed as commit date minus author time. This patch does that, creating a new field in the enriched index.

Uploading this on behalf of @BenLloydPearson.

This should close #235 

We're uploading this without proper testing. Please, have a detailed look before accepting.

### DIFF
--- a/grimoire_elk/elk/git.py
+++ b/grimoire_elk/elk/git.py
@@ -343,6 +343,8 @@ class GitEnrich(Enrich):
         eitem["utc_author"] = (author_date - author_date.utcoffset()).replace(tzinfo=None).isoformat()
         eitem["utc_commit"] = (commit_date - commit_date.utcoffset()).replace(tzinfo=None).isoformat()
         eitem["tz"] = int(author_date.strftime("%z")[0:3])
+        # Compute time to commit
+        eitem["time_to_commit"] = (eitem["utc_commit"] - eitem["utc_author"]).hours()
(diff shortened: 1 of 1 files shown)


acs (approved):
LGTM!

PR #239, opened 2018-03-08
[arthur] Rename arthur file to elk

valeriocos:
This patch renames the name of the packages ocean and elk and the name of arthur.py. It also modifies the import sections of many files in the repository


acs (commented):
arthur is used also in mordred and not sure if in manuscripts.  I love the rename, but I will do it after we release 18.03-01

acs:
After thinking more about it, arthur.py does much more than just feeding. It has the core logic of GrimoireELK so probably the best name is just `elk.py`. What do you think @valeriocos ?

valeriocos:
elk.py looks good to me. However in this case we should rename the sub-package **elk** with something else like **enrich** (to avoid misunderstandings between the new elk.py and the package). I would rename also the package **ocean** with a new name (for instance **raw**).

acs (approved):
LGTM

PR #240, opened 2018-03-08
Gender

alpgarcia:
I did 3 separate commits:
* Add gender info: I followed the most simple approach. If there is gender info coming from sortinghat, add that info to the enriched index. Future work could add a parameter to ignore this information. Current code asumes we want that info if available in sortinghat.
* Add tests for gender methods: basically two test for the methods I've modified. As I see those as a whole, I tested all cases I was aware of and tests started to fail. I decided not to fix the methods in this commit to make clear the errors and the fixes needed (next commit)
* Fix errors mentioned above: this allow to review particular fixes not related with gender info. Imho, they are needed to have a more robust code.

Thanks!

### DIFF
--- a/grimoire_elk/elk/enrich.py
+++ b/grimoire_elk/elk/enrich.py
@@ -595,10 +595,10 @@ class Enrich(ElasticItems):
         if identity:
             # Use the identity to get the SortingHat identity
             sh_ids = self.get_sh_ids(identity, self.get_connector_name())
-            eitem_sh[rol + "_id"] = sh_ids['id']
-            eitem_sh[rol + "_uuid"] = sh_ids['uuid']
-            eitem_sh[rol + "_name"] = identity['name']
-            eitem_sh[rol + "_user_name"] = identity['username']
+            eitem_sh[rol + "_id"] = sh_ids.get('id', '')
+            eitem_sh[rol + "_uuid"] = sh_ids.get('uuid', '')
+            eitem_sh[rol + "_name"] = identity.get('name', '')
+            eitem_sh[rol + "_user_name"] = identity.get('username', '')
             eitem_sh[rol + "_domain"] = self.get_identity_domain(identity)
         elif sh_id:
             # Use the SortingHat id to get the identity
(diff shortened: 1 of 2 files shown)


grimoire_elk/elk/enrich.py
```diff
@@ -595,10 +595,14 @@ def get_item_sh_fields(self, identity=None, item_date=None, sh_id=None,
         if identity:
             # Use the identity to get the SortingHat identity
             sh_ids = self.get_sh_ids(identity, self.get_connector_name())
-            eitem_sh[rol + "_id"] = sh_ids['id']
-            eitem_sh[rol + "_uuid"] = sh_ids['uuid']
-            eitem_sh[rol + "_name"] = identity['name']
-            eitem_sh[rol + "_user_name"] = identity['username']
+            if 'id' in sh_ids:
```
sduenas: I would use the `get` method  of `dict` type instead of a set of  `if` blocks. For example:

```
eitem_sh[rol + "_id"] = sh_ids.get('id', None)
```

tests/test_enrich.py
```diff
@@ -0,0 +1,378 @@
...
+            return vals[uuid]
+        self._enrich.get_unique_identity = MagicMock(side_effect=side_effect)
+
+        profile = self._enrich.get_profile_sh('00000')
+        self.assertEqual(profile['name'], uidentity.profile.name)
+        self.assertEqual(profile['email'], uidentity.profile.email)
+        self.assertEqual(profile['gender'], uidentity.profile.gender)
+        self.assertEqual(profile['gender_acc'], uidentity.profile.gender_acc)
+
+    def test_get_item_sh_fields(self):
```
sduenas: I think you should split this method in several tests that will run independently. The idea is to test as much as possible cases at the same time.

sduenas (changes requested):

alpgarcia:
I've addressed both comments. I've also changed the code to ensure gender fields are always included in the enriched index and are not empty. `Unknown` and `0` values are used for `author_gender` and `author_gender_acc` instead.

sduenas (approved):

acs (approved):
LGTM

PR #242, opened 2018-03-12
[mappings] Temporally revert mappings

valeriocos:
This patch temporally reverts mappings for a small subset of data sources to allow the correct creation of the corresponding dashboards.


acs (approved):
LGTM. Time to return to previous version so Kibana panels work and then progressively, we will fix the mappings.

Issue #221, opened 2018-02-23
Removal of the code that generates the "conf" index


hmitsch:
**Desirement Source:** 
https://github.com/mozilla/participation-metrics-org/issues/157

**Short Description:**
A vulnerability assessment conducted by [Mozilla's Enterprise Information Security](http://infosec.mozilla.org/) team revealed, that
> https://analytics.mozilla.community/elasticsearch/conf/_search leaks some configuration data eg:
"git command - fatal: could not create leading directories of '/home/bitergia/.perceval/repositories/https://github.com/MozillaFoundation/Design.git-git'\n"

**Potential Solution:**
According to @acs:
> This data is not used anymore. Is legacy stuff that must be removed. 

/cc @gdestuynder @sanacl

acs:
@hmitsch once https://github.com/chaoss/grimoirelab-elk/pull/243 gets accepted the index won't be created anymore. And it is safe to remove it from the current deployments.

valeriocos:
better late than never, closing this issue since fixed by #243

PR #243, opened 2018-03-14
Remove conf index

acs:
This index is not used anymore (for years now) and it contains sensible information. Its removal has been asked by https://github.com/chaoss/grimoirelab-elk/issues/221

Once this PR gets accepted, the removal will be completed.

Old scripts not used anymore that depends on this index have been removed also.


valeriocos (approved):
LGTM

PR #245, opened 2018-03-14
[ocean] Add ES 2.x mappings for ocean backends

valeriocos:
This patch includes ElasticSearch 2.x mappings for several ocean backends. In some cases, the patch also modifies the if condition used to detect the ElasticSearch version


acs (approved):
LGTM

PR #246, opened 2018-03-14
[enrich][bugzillarest] Change " " with "+" in product and component names to follow Mozilla projects mapping

acs:
The mapping definition for projects is done for bugzillarest like:

```
    "Localization": {
        "bugzillarest": [
....
            "https://bugzilla.mozilla.org/buglist.cgi?product=Mozilla+Localizations&component=bg+/+Bulgarian",
```

White spaces are changed with "+" because the name of the component in the raw items from Bugzilla is: `"component" : "bg / Bulgarian"`. It is done in all cases like that way: only changing " " with "+" (but it is not a urlencode).

This PR implement that.

### DIFF
--- a/grimoire_elk/elk/bugzillarest.py
+++ b/grimoire_elk/elk/bugzillarest.py
@@ -26,6 +26,7 @@
 import logging
 
 from datetime import datetime
+from urllib.parse import quote_plus
 
 from dateutil import parser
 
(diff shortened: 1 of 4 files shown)


grimoire_elk/elk/bugzillarest.py
```diff
@@ -110,15 +111,15 @@ def get_item_project(self, eitem):
...
 
-        component = eitem['component']
         url = eitem['origin']
-        product = eitem['product']
+        component = eitem['component'].replace(" ", "+")
+        product = eitem['product'].replace(" ", "+")
 
         repo_comp = url + "/buglist.cgi?product=" + product + "&component=" + component
-        repo_comp1 = url + "/buglist.cgi?component=" + component + "&product=" + product
+        repo_comp_prod = url + "/buglist.cgi?component=" + component + "&product=" + product
```
valeriocos: I don't understand the difference between **repo_comp** and **repo_comp_prod**, should the order of the parameters in the URL be independent?

valeriocos (commented):
It looks good to me, I have just a question on the variables repo_comp and repo_comp_prod

grimoire_elk/elk/bugzillarest.py
acs: We have the two cases (first product and then the component, and the other case) depending in how the mapping is specified. This is why I checked both. But `repo_comp` must be `repo_prod_comp` following the same approach.

valeriocos (approved):
LGTM

PR #247, opened 2018-03-15
[askbot] Add information to track accepted answers

valeriocos:
This code adds to the enriched item information about: (i) whether a question has an accepted answer, (ii) the id of the accepted answer


acs (approved):
LGTM

PR #253, opened 2018-03-16
[ocean] Prevent Slack attachment info to be indexed

valeriocos:
This patch avoids to define dynamic mappings for attachment data. Thus it prevents number format exceptions related to the field data.attachments.ts (which is not recognized as number)


acs (approved):
LGTM

PR #254, opened 2018-03-16
[ocean/elk] Limit content size of keyword fields

valeriocos:
This patch limits the content size of keyword fields to 32000, thus it avoids (i) losing items in the raw and enriched indexes (ii) breaking the panel specifications


acs (commented):
The problem with this approach is what happens if we want the full message. IMHO the problem is bucketing in Kibana using fields like this one.

If we decide to follow this approach, I will have two fields:

```
eitem['message'] = commit['message'][:256]
eitem['message_full'] = commit['message']
```
and define 256 as a global constant, but with this smaller value.

valeriocos:
The full message is stored in the raw index, so at any moment it can be added to the enriched index. However, I'm not sure the panel uses the full message, so I wouldn't add another index.

acs (approved):
LGTM

PR #255, opened 2018-03-19
[ocean] Add ES 5.1.x mappings for ocean backends

valeriocos:
This patch includes ElasticSearch 5.1.x mappings for several ocean backends


acs (approved):
LGTM. In ES 5.1 if a mapping starts with tab it is not added.

PR #256, opened 2018-03-19
[elk] Fix check condition ES version for meetup, nntp, rss

valeriocos:
This patch fixes how the ES version is checked. The parameter of get_elastic_mappings (es_major) is checked against a string (not an integer).


acs (approved):
LGTM

Issue #257, opened 2018-03-19
eitem['description'][:self.KEYWORD_MAX_SIZE] fails if it is None


acs:
During the mordred testing an error has appeared:

```
mordred_1        |   File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/elk/redmine.py", line 135, in get_rich_item
mordred_1        |     eitem['description'] = eitem['description'][:self.KEYWORD_MAX_SIZE]
mordred_1        | TypeError: 'NoneType' object is not subscriptable
```

Before truncating the fields it is needed to test if they have a value.

PR #258, opened 2018-03-20
[elk] Fix limit description on none values for redmine

valeriocos:
This patch targets #257. It fixes the truncation of descriptions during the enrichment of redmine items.


acs (approved):
LGTM

PR #260, opened 2018-03-21
[ocean] Fix mappings with jenkins to avoid regression errors

valeriocos:
This patch fixes the mappings for jenkins in order to avoid conflicts with the existing mapping on data.changeSet.items.comment.


acs (approved):
LGTM

PR #262, opened 2018-03-21
[elk] Create areas of code output index if needed

alpgarcia:
If areas of code output index does not exist and incremental mode
is enabled, then the index is created.

Notice that index names have been modified to follow a similar convention as onion study. Once #261 gets accepted, I will update documentation for this study to the MD file created in that PR.


acs (approved):
LGTM

PR #263, opened 2018-03-21
[arthur] Fix params parsing for arthur tasks

acs:
All perceval backends are now supported in arthur except the ones provided by perceval-mozilla/opnfv/puppet because and problem installing their pip packages with arthur.

### DIFF
--- a/grimoire_elk/ocean/dockerhub.py
+++ b/grimoire_elk/ocean/dockerhub.py
@@ -92,3 +92,13 @@ class DockerHubOcean(ElasticOcean):
         params = url.split()
 
         return params
+
+    @classmethod
+    def get_arthur_params_from_url(cls, url):
+        # In the url the org and the repository are included
+
+        params = url.split()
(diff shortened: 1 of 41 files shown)


valeriocos (commented):
should we provide some tests for this code?

acs:
I have doubts about adding tests to this code. Probably yes. But this code is working against several pieces (arthur, config from mordred ...) so I am not really sure. But it is critical code that joins arthur and mordred so ... yes, let's add tests! I will do that. Thanks @valeriocos! I will update the PR. Also, we need to think about the doc related to it. But the doc for all this params is in perceval already.

acs:
@valeriocos tests added for all data sources! Could you review them? If all is ok, please merge directly this PR! Thanks :+1:

valeriocos:
@acs, please fix the flake8 errors

acs:
flake8 issues fixed!

valeriocos:
LGTM

PR #264, opened 2018-03-22
[ocean] Add ES 2.x mappings for askbot backend

valeriocos:
This code enables ES 2.x mappings for the askbot backend.


acs (approved):
LGTM

PR #266, opened 2018-03-22
[elk] Include mappings to discourse backend

valeriocos:
This code enables mappings for the enrichment phase of the discouse backend.


acs (approved):
LGTM

PR #267, opened 2018-03-22
[doc] Add documentation for Areas of Code study

alpgarcia:
I've included information about index fields being generated. Ideally this should be changed by a link to the corresponding CSVs.

I've also added a broken link to the place where Mordred config documentation should be, please check if this is the right place.


acs (approved):
LGTM. Great job @alpgarcia

PR #270, opened 2018-03-26
[raw] Fix typos in ES 2.2 mappings for gerrit

valeriocos:
This patch fixes the mappings for ES 2.2 defined for gerrit data.


acs (approved):
LGTM

PR #272, opened 2018-03-26
[enrich] Fix limit comment message in gerrit

valeriocos:
This patch checks that comments within a review exist, before limiting their size.


acs (approved):
LGTM

PR #273, opened 2018-03-26
[enriched] Set not accepted answer flag to askbot unanswered question

valeriocos:
This patch sets the default value of the attribute question_has_accepted_answer to zero for unanswered questions in askbot.


acs (approved):
LGTM

PR #276, opened 2018-03-26
[enriched] Handle wrong time offset in Git enrichment

valeriocos:
This fix handles wrong time offsets of authored and committed dates. Thus, it prevents Value errors related to parsing wrong dates.


acs (approved):
LGTM

Issue #275, opened 2018-03-26
[askbot] Add `answer_status` field


alpgarcia:
Type: Keyword
Values: ['Accepted', 'Not Accepted']

This field should be stored only for **answers**, Questions and comments must not store it.

alpgarcia:
Please, update CSV file before closing this:
https://github.com/chaoss/grimoirelab-elk/blob/master/schema/askbot.csv

alpgarcia:
Solved by #287

PR #278, opened 2018-03-26
[enrich] Add `answer_status` field to askbot

valeriocos:
This PR targets issue #275. Thus the proposed code includes a new attribute in the enriched item, used to track the status of answers of a given question.


acs (approved):
LGTM

PR #286, opened 2018-03-28
[studies] Add timeout to ES connections

alpgarcia:
Default timeout is set to 10, now is explicitely set to 100 to
avoid failures when running studies through a slow network
connection.


acs (approved):
LGTM

PR #287, opened 2018-03-28
[schema] Update askbot.csv

valeriocos:
This patch adds a new keyword field in the askbot.csv, which gives insights about the status of answers (i.e., accepted or not_accepted).


alpgarcia:
LGTM, thanks!!

PR #290, opened 2018-04-03
[enrich] Handle missing accepted_answer_id attribute

valeriocos:
This patch initializes the attribute accepted_answer_id to a null value when it is not present in the json returned by the api. Thus, it avoids KeyError exceptions during the enrichment phase.


acs (approved):
LGTM

PR #291, opened 2018-04-03
[raw] Change discourse mapping for data.post_stream.posts.cooked

valeriocos:
This patch enables to map the attribute data.post_stream.posts.cooked to a text type. Thus, it avoids immense term exceptions which may occur when using the keyword type.


acs (approved):
LGTM

PR #292, opened 2018-04-03
[enrich] Fix mapping for attribute question_title in discourse

valeriocos:
This patch fixes the mapping of the enriched attribute question_title within the discourse backend. It allows the attribute to be searchable and aggregatable, thus avoiding breaking the corresponding panel.


acs (approved):
LGTM

PR #293, opened 2018-04-04
[enrich] Limit size of value attribute summary in bugzilla

valeriocos:
This patch limits the size of the attribute summary to 30000 to avoid immense term errors when storing enriched items to ElasticSearch.


acs (approved):
LGTM

PR #294, opened 2018-04-04
[enrich] Limit size of summary and main_description in bugzillarest

valeriocos:
This patch limits the size of the attributes summary and main_description to avoid immense terms error when storing enriched items to ElasticSearch.


acs (approved):
LGTM

acs (approved):
LGTM

PR #295, opened 2018-04-04
[enrich] Fix flake8 error in discouse

valeriocos:
This patch fixes a flake8 error (not used import) in the discourse backend.


acs (approved):
LGTM

PR #296, opened 2018-04-04
[tests] Update test data for Jira

valeriocos:
This code updates the data used to test the jira raw/enrich backends. In particular, the description within data.renderedFields is inflated in order to check whether the mapping defined in the raw backend properly works.


acs (approved):
LGTM

PR #297, opened 2018-04-04
[tests] Update tests data for Slack

valeriocos:
This code updates the data used to test the slack raw/enrich backends. In particular, attachments information is added to check whether the mapping defined in the raw backend properly works.


acs (approved):
LGTM

PR #298, opened 2018-04-04
[raw] Update mappings to filter body information in confluence items

valeriocos:
This patch prevents to index body information which is not used in panels (as reported in the /schema/discourse.csv). Test data has been evolved accordingly to check that the new mapping properly
works.


acs (approved):
LGTM

PR #299, opened 2018-04-06
[enrich] Handle revisions without comment in mediawiki

valeriocos:
This patch allows to handle revisions that have no comment. Thus, it avoids KeyError exception during the enrichment phase.


acs (approved):
LGTM

PR #300, opened 2018-04-06
[enrich] Handle different formats for failures in functest

valeriocos:
This patch allows to handle the different formats for failures in functest. In a nutshell, failures are returned by the functest API as number or list. This heterogeneity causes ES to fail since it cannot map different types for the same attribute. Thus, this code stores the lenght of the list in the attribute failures, avoiding type conflicts.


acs (approved):
LGTM

PR #301, opened 2018-04-07
[raw] Filter attachment-related data of slack messages

valeriocos:
This patch prevents to index attachment-related information which may cause timely errors (due to the timestamp value not always recognized as such by ES). This data is not used in the corresponding panel, as shown in the schema/slack.csv.


acs (approved):
LGTM

PR #302, opened 2018-04-09
[raw/enrich] Handle long description in Jira fields

valeriocos:
This PR allows to handle long fields.description values in the raw and enriched indexes. In the raw index, a mapping is defined to to avoid immense term exceptions, while in the enriched index, the value is truncated to avoid changing the type of the attribute (keyword).


acs (approved):
LGTM

PR #306, opened 2018-04-10
[enrich] Handle long text values in  mozillaclub

valeriocos:
This code allows to handle long text values without breaking the corresponding panel. Thus, the previous mappings defined to treat long text values as text type have been changed to handle the
same information as keyword type. However, in order to avoid immense term exceptions which can occur with the keyword type, the text values are truncated to 30000 chars.


acs (approved):
LGTM

PR #309, opened 2018-04-11
[enrich][metadata] Add project metadata fields to the enriched items

acs:
The project fields that are added to the enriched items when there is
project information now includes also the meta fields if they exits.


valeriocos (commented):
LGTM

valeriocos (approved):

PR #310, opened 2018-04-12
[nntp] Use mbox mappings for NNTP items

valeriocos:
This code modifies the mappings for the NNTP backend, which are now the same of the mbox backend. Tests have been evolved accordingly.


acs (approved):
LGTM

PR #311, opened 2018-04-12
[mbox/nntp] Remove X-fields in raw data    

valeriocos:
This patch removes from the raw data the custom fields (X-...) to avoid hitting the limit of 1000 fields in ES.


acs (approved):
LGTM

PR #312, opened 2018-04-12
[phabricator] Handle dots in JSON attribute names

valeriocos:
This patch replaces dots in field attribute names with underscores. Thus, it prevents errors on ES versions previous to 5 (which don't allow to store this kind of attributes). For consistency reasons, this fix is applied also to more recent versions of ES.


acs (approved):
LGTM

PR #314, opened 2018-04-12
[tests] Add test for load identities in sortinghat

valeriocos:
This code includes a test to measure the time used to retrieve and store identities in sortinghat via elk. This test can be useful to measure and improve the performance of elk/sortinghat when loading identities.


acs (approved):
LGTM

acs:
I will merge it once the release is done!

acs:
More changes in gelk are needed before the release, so adding this change now.

PR #317, opened 2018-04-12
Add alias management to studies

alpgarcia:
These commits add functionality to manage aliases after onion and areas of code execution. It is particularly important for onion because each time the study runs, index is deleted and therefore the corresponding alias is automatically removed. This way, once data is available, alias is created if needed.


acs (approved):
LGTM

PR #318, opened 2018-04-13
[identities] Load identities to sortinghat by bulks

valeriocos:
This PR changes the way of loading identities to sortinghat. In a nutshell, instead of loading in memory all the identities before sending them to sortinghat, they are sent to bulks of 500. Thus, the consuption of memory is safe when dealing with a large number of identities. Furthermore, this PR enables also to log exceptions that may occur when processing identities. This log can help to shed some light on mordred
threads that from time to time die without leaving any trace.

### DIFF
--- a/grimoire_elk/elk.py
+++ b/grimoire_elk/elk.py
@@ -38,6 +38,7 @@ from perceval.backend import find_signature_parameters, Archive
 
 from .utils import get_elastic
 from .utils import get_connectors, get_connector_from_name
+from .enriched.sortinghat_gelk import SortingHat
 from .enriched.utils import get_last_enrich, grimoire_con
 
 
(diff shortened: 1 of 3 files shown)


grimoire_elk/elk.py
```diff
@@ -381,18 +382,31 @@ def load_identities(ocean_backend, enrich_backend):
...
+
             logger.debug("Processed %i items identities (%i identities) from %s",
                          items_count, len(new_identities),
                          enrich_backend.get_connector_name())
-    logger.debug("TOTAL ITEMS: %i", items_count)
+            new_identities = []
 
-    logger.info("Total new identities to be checked %i", len(new_identities))
+    if new_identities:
+        identities_count += len(new_identities)
```
sduenas: This code is exactly the same you wrote above. Don't repeat yourself. Put it inside a method or function.

sduenas (changes requested):

sduenas (approved):

acs:
After talking with @valeriocos there are no performance drawbacks adding the identities by bulks, and it is more robust (you don't need to have all 1 million identities loaded in memory to start doing things) so I merge this PR: Thanks!

PR #319, opened 2018-04-16
[jira] Handle data.changelog.histories fields

valeriocos:
This patch prevents to index changelog histories of Jira items, thus avoiding immense term exceptions when inserting data to ES.


acs (approved):
LGTM

PR #321, opened 2018-04-17
[jenkins] Update raw mappings to deal with large comments

valeriocos:
This patch allows to deal with large comments of Jenkins data, thus it avoids immense term exceptions when inserting data to ES.


acs (approved):
LGTM

PR #324, opened 2018-04-19
[phabricator] Handle long description.raw value in raw items

valeriocos:
This patch prevents to index description.raw attributes of phabricator items, thus avoiding immense term exception in ES. Description.raw attribute is not used neither in the enriched item nor in the corresponding panels (according to the csv in the schema folder).


acs (approved):
LGTM

PR #325, opened 2018-04-24
[enrich] Remove emails from all enriched indexes

acs:
Emails were detected in git, github, jira and mbox. This commit
removes all the fields including emails.

### DIFF
--- a/grimoire_elk/enriched/git.py
+++ b/grimoire_elk/enriched/git.py
@@ -332,14 +332,14 @@ class GitEnrich(Enrich):
         self.__fix_field_date(commit, 'CommitDate')
 
         # data fields to copy
-        copy_fields = ["message", "Author"]
+        copy_fields = ["message"]
         for f in copy_fields:
             if f in commit:
                 eitem[f] = commit[f]
             else:
                 eitem[f] = None
         # Fields which names are translated
-        map_fields = {"commit": "hash", "message": "message_analyzed", "Commit": "Committer"}
+        map_fields = {"commit": "hash", "message": "message_analyzed"}
         for fn in map_fields:
             if fn in commit:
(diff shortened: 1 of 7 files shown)


jgbarah:
> Some of the fields with emails are included in the CSV files which define the fields needed by the panels:

AFAIK, the CSV files include *all* fields in the current indexes, not only those used by the panels. Please check with @alpgarcia and @dicortazar.

jgbarah (commented):
Shoudln't it better to do it the other way around? I mean, checking all fields that have not a real reason to be in the index, and then remove them (not only those with email addresses).

Obviously, just a comment. maybe could just be the origin for a new issue, while this pr is accepted.

acs:
> Shoudln't it better to do it the other way around? I mean, checking all fields that have not a real reason to be in the index, and then remove them (not only those with email addresses).

Yes, but this is another issue that should be addressed in other ticket.

acs:
> AFAIK, the CSV files include all fields in the current indexes, not only those used by the panels. Please check with @alpgarcia and @dicortazar.

Yes, you are right. The fields in the CSV were the fields that appeared in the mapping of enriched indexes, not the fields used by panels.

valeriocos:
I guess the best solution is to create some samples for the enriched backends modified by this PR and try them on blue5, to be sure that no panel will get broken. 
The sample indexes are in the process of being created, then they will be uploaded to blue5.

valeriocos:
The PR doesn't break the panels.
Should the csv files being updated to reflect the new attributes in the enriched indexes?

acs:
> Should the csv files being updated to reflect the new attributes in the enriched indexes?

Yes!

valeriocos (commented):
LGTM

PR #327, opened 2018-04-26
[jira] Disable dynamic mappings for data.changelog.histories

valeriocos:
This patch disable dynamic mappings when indexing changelog histories of Jira items.


acs:
LGTM

PR #328, opened 2018-04-26
[tests] Temporary disable tests using CSVs

valeriocos:
This patch disables the tests using the CSVs which are currently failing due to not updated/not reviewed CSVs.


acs (approved):
LGTM

PR #329, opened 2018-04-26
[mappings] Goodbye ES 2.x

valeriocos:
This PR removes the code tailored to work with raw and enriched indexes in ES 2.x.
The PR has been splitted in several commits, one for data source, to simplify its reviewing. 
Note that this PR should be merged after https://github.com/chaoss/grimoirelab-elk/pull/328 to avoid failures with the tests relying on CSVs.


acs:
LGTM

PR #331, opened 2018-04-27
[elastic] Remove not used global_mapping method and limit the keyword fields

acs:


valeriocos (commented):
I would remove only the global_mapping method.
The use of the ignore_above can be risky (https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html). Wrt KEYWORD_MAX_SIZE, a similar variable is already defined enriched/enrich.py

acs:
> The use of the ignore_above can be risky (https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html)

Guau, I expected that they were cut to the limit, not that they will not be indexed or stored. They are using this feature in the "string" automatic mapping, so I supposed that they were cutting the string. My fault. Thanks for the catch @valeriocos 

> Wrt KEYWORD_MAX_SIZE, a similar variable is already defined enriched/enrich.py

Yes, I know it. But if we have used the PR approach, it should be here. So this change must not be added. 

I will rework now the PR with only the method removal.

acs:
PR updated!

PR #338, opened 2018-05-03
[stackexchange] Add fields to csv to track accepted answers

valeriocos:
This code add new fields to stackoverflow.csv about:
(i) whether a question has an accepted answer,
(ii) the id of the accepted answer.

### DIFF
--- a/schema/stackoverflow.csv
+++ b/schema/stackoverflow.csv
@@ -1,4 +1,5 @@
 name,type
+answer_status,keyword
 answer_count,long
 answer_id,long
 answer_tags,keyword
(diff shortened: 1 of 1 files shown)


schema/stackoverflow.csv
```diff
@@ -1,5 +1,7 @@
-name,type
+name,type
+answer_status,keyword
 answer_count,long
+answer_status
```
alpgarcia: I think this line needs to be removed :)

alpgarcia (changes requested):

alpgarcia (approved):

Issue #336, opened 2018-04-30
[Stackexchange] Issue getting user['display_name']


jsmanrique:
Analyzing [Ubuntu StackExchange instance](https://askubuntu.com/) with Grimoire Mordred, I get the following error message:
```
Collection for stackexchange: finished after 01:13:51 hours
Enrichment for stackexchange: starting...
2018-04-30 10:07:13,609 - grimoire_elk.elk - ERROR - Traceback (most recent call last):
  File "/home/jsmanrique/grimoirelab/venv/lib/python3.5/site-packages/grimoire_elk/elk.py", line 583, in enrich_backend
    total_ids = load_identities(ocean_backend, enrich_backend)
  File "/home/jsmanrique/grimoirelab/venv/lib/python3.5/site-packages/grimoire_elk/elk.py", line 380, in load_identities
    identities = enrich_backend.get_identities(item)
  File "/home/jsmanrique/grimoirelab/venv/lib/python3.5/site-packages/grimoire_elk/enriched/stackexchange.py", line 108, in get_identities
    user = self.get_sh_identity(answer[identity])
  File "/home/jsmanrique/grimoirelab/venv/lib/python3.5/site-packages/grimoire_elk/enriched/stackexchange.py", line 90, in get_sh_identity
    identity['username'] = user['display_name']
KeyError: 'display_name'

2018-04-30 10:07:13,637 - grimoire_elk.elk - ERROR - Error enriching ocean from stackexchange (askubuntu.com): 'display_name'
Elasticsearch aliases for stackexchange: creating...
Elasticsearch aliases for stackexchange: created!
```
It seems it's expecting a `user['display_name']` and it's empty or it doesn't even exists..

acs:
@valeriocos could you take a look?

PR #343, opened 2018-05-03
[stackexchange] Handle deleted users

valeriocos:
This PR targets https://github.com/chaoss/grimoirelab-elk/issues/336. It allows to handle deleted users on stackexchange platforms, which displayed names are now replaced with 'unknown' to avoid sortinghat failing when loading identities.

### DIFF
--- a/grimoire_elk/enriched/stackexchange.py
+++ b/grimoire_elk/enriched/stackexchange.py
@@ -76,6 +76,9 @@ class StackExchangeEnrich(Enrich):
             # for answers
             user = item[identity_field]
 
+        if 'display_name' not in user:
+            user['display_name'] = ''
+
         identity['username'] = user['display_name']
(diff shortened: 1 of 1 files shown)


grimoire_elk/enriched/stackexchange.py
```diff
@@ -76,6 +76,9 @@ def get_sh_identity(self, item, identity_field=None):
             # for answers
             user = item[identity_field]
 
+        if 'display_name' not in user:
+            user['display_name'] = 'unknown'
```
acs: In similar situations in the past, we are just filling the entry with '' (empty string). If we decide to use a string like 'unknown' it should be a constant defined in enriched base class a start using it in other places. IMHO in this case, I will use an empty string.

acs (approved):
LGTM

PR #344, opened 2018-05-03
[twitter] Limit number of indexed fields

valeriocos:
This code limits the number of fields indexed when inserting raw data in ES. Although a csv file/schema for twitter is not available, those fields should not be used in the corresponding panel


acs (approved):
LGTM

PR #345, opened 2018-05-03
[twitter] Change date value to index tweets

valeriocos:
This code changes the date value used to index tweets. Tweets come from logstash and the API, both include a @timestamp attribute, which is now used to index them.


acs:
@valeriocos could you check the error in tests?

dpose:
LGTM!

acs:
LGTM

PR #348, opened 2018-05-09
[elk] raise exception in do_studies

valeriocos:
This patch enables to raise an exception when a study fails, instead of printing the traceback. Thus, it allows Mordred to be aware that a thread has failed.


valeriocos:
This PR should be reviewed together with https://github.com/chaoss/grimoirelab-sirmordred/pull/144.

acs (approved):
LGTM

PR #352, opened 2018-05-17
[enrich][twitter] Use case insesitive tags for doing the projects mapping

acs:


valeriocos (approved):
LGTM. Instead of using the function .lower(), another solution could be to use the CaseInsesitiveDict in requests library (requests.structures import CaseInsensitiveDict)

acs:
> LGTM. Instead of using the function .lower(), another solution could be to use the CaseInsesitiveDict in requests library (requests.structures import CaseInsensitiveDict)

Great suggestion. I have reimplemented it using this class. Now it is case insensitive for both places: projects mapping definition and twitter items. Thanks @valeriocos !

PR #354, opened 2018-05-22
[elk] Fix enrichment of stackexchage answers

valeriocos:
This code fixes the enrichement of stackexchange questions, which now generates items that include the attributes `is_accepted_answer` and `answer_status`.


acs (approved):
LGTM

PR #355, opened 2018-05-23
[raw] Reduce the number of fields indexed for jenkins items

valeriocos:
This code enhances the raw mappings for jenkins data to avoid `exceed limit number of fields` exceptions.


acs (approved):
LGTM

PR #356, opened 2018-05-24
[elk] Propagate study params from mordred to elk

valeriocos:
This code allows to propagate the params defined in the mordred configuration files to elk. Depending on the study, elk now is able to set up the parameters for the study. The studies that benefit of this patch are onion, areas of code and demography.


valeriocos:
This PR should be reviewed together with https://github.com/chaoss/grimoirelab-sirmordred/pull/150

acs (approved):
LGTM

PR #357, opened 2018-05-24
[gelk_mapping.py] Script to show the mappings for a data source

acs:


valeriocos (commented):
Please, @acs fix the flake8 errors. It is possible to include some tests for the code you are proposing?

acs:
> Please, @acs fix the flake8 errors. It is possible to include some tests for the code you are proposing?

Done!

About the testing, I am not sure we should add testing in such simple scripts and the best way to do it. If we decide to include tests, I would just include directly inside the script file, the unit testing classes. What do you think @valeriocos ?

jgbarah:
I see I'm late to this review. But just in case: there is no change to setup.py to include this new script, which means it will not be included in the package. @acs could you please produce a new pr with that?

acs:
> I see I'm late to this review. But just in case: there is no change to setup.py to include this new script, which means it will not be included in the package. @acs could you please produce a new pr with that?

Done!

PR #358, opened 2018-05-25
[gelk_mapping] Generate files with the JSON mappings to be used

acs:
Generate files with the JSON mappings to be used by elasticdump to create the indexes


valeriocos (approved):
LGTM

PR #362, opened 2018-05-30
[twitter] Set field_date for twitter enricher

valeriocos:
This code changes the field_date to index twitter data from `@timestamp` to `created_at`


acs (approved):
LGTM

PR #365, opened 2018-06-05
[enrich] Modify max length for keyword fields

valeriocos:
This code sets to 30000 the max length for fields of type keyword, thus preventing immense term exceptions during the creation of enriched items.


acs (approved):
LGTM

PR #370, opened 2018-06-06
[studies][demography] Use author_uuid field to aggregate the authors instead of author_id

acs:
And author could have several author_id fields so it can not be used to aggregate authors.

### DIFF
--- a/grimoire_elk/enriched/git.py
+++ b/grimoire_elk/enriched/git.py
@@ -668,7 +668,7 @@ class GitEnrich(Enrich):
           "aggs": {
             "author": {
               "terms": {
-                "field": "author_id",
+                "field": "author_uuid",
                 "size": 10000
               },
               "aggs": {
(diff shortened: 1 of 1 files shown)


valeriocos (approved):
LGTM

aswanipranjal:
@valcos, I had a doubt about `author_uuid` field. When I am enriching a git index using the `p2o.py` script, there is no filed named `author_uuid` in the enriched index created.

valeriocos:
@aswanipranjal those fields are included when executing sortinghat (it may be possible that you are executing `p2o` without sortinghat?). The demography study (if on) is launched after the enrichment, and includes some extra fields such as `author_min_date` and ' author_max_date`.

aswanipranjal:
Ah, yes! I was not using sorting hat! Thank you very much @valeriocos!!

valeriocos:
you are welcome @aswanipranjal

PR #371, opened 2018-06-06
[stackexchage] Fix fields for questions having an accepted answer

valeriocos:
This patch fixes a bug that was causing to not properly fill the fields `question_accepted_answer_id` and `question_has_accepted_answer` for questions having only one answer (which was marked as accepted).


acs (approved):
LGTM

Issue #340, opened 2018-05-03
[studies] No method `latest_date()` in `Connector` class interface 


alpgarcia:
`latest_date()` method is missing from `Connector` class:

https://github.com/chaoss/grimoirelab-elk/blob/f350511b4904b1335e9d9a62c8956d32e97ccd6e/grimoire_elk/enriched/ceres_base.py#L80

PR #372, opened 2018-06-11
[enriched] Add `latest_date` method to connector

alpgarcia:
`latest_date` is used by CeresBase but was not included in the
definition Connector abstract class.

This is not a high priority bug as it does not affect to code execution.

Solves #340


valeriocos (approved):
LGTM

PR #373, opened 2018-06-12
[raw] Update mappings for jenkins to ES 5.x

valeriocos:
This code updates the mappings for Jenkins to make them compatible with versions of ES greater or equal to 5.x.


acs:
LGTM

PR #374, opened 2018-06-12
[elastic] Add template to force date format

valeriocos:
This code allows to force the format of date fields. Thus, those fields are now formatted according to `strict_date_optional_time||epoch_millis`.


acs (approved):
Not sure if we really need that, but approving it.

PR #375, opened 2018-06-12
[enrich] Handle deleted user in stackoverflow

valeriocos:
This code allows to deal with deleted users in stackoverflow questions, logging a warning to notify platform operator.


acs (approved):
LGTM

PR #376, opened 2018-06-18
[bugzilla] Fix immense term exception for activity attribute

valeriocos:
This code avoids indexing fields within the activity attribute, thus avoiding immense term exception when inserting raw data to ElasticSearch.


acs (approved):
LGTM

PR #377, opened 2018-06-18
[jira] Fix mapping to avoid limit number of fields exception

valeriocos:
This code reduces the number of indexed fields for jira raw indexes, thus avoiding limit number of fields exceptions thrown by ES when inserting new data.


acs (approved):
LGTM

PR #383, opened 2018-06-24
Add code to calculate first response to issue duration

### LINKED ISSUE #8: GMD: first response to issue duration
Following #1, the definition states: *Time between a new issue is opened and a maintainer responds*. 
- Who is considered to be a maintainer? In the raw index, the users have 5 types of author_association: None, Owner, Collaborator, Contributor and Member. So are the latter 4 the maintainers for that project?
- What can be considered as a response? A reaction or a comment?

metric: [first response to issue duration](https://github.com/chaoss/metrics/blob/master/activity-metrics/first-response-to-issue-duration.md).

aswanipranjal:
closes chaoss/wg-gmd#8

This PR adds the functionality to calculate the first response to issue duration for an issue/pr by someone other than the user who created the issue/pr and adds it to the enriched index for github data source.

### DIFF
--- a/grimoire_elk/enriched/github.py
+++ b/grimoire_elk/enriched/github.py
@@ -28,6 +28,7 @@ import logging
 import re
 
 from datetime import datetime
+from dateutil import parser
 
 from .utils import get_time_diff_days
 
@@ -249,6 +250,20 @@ class GitHubEnrich(Enrich):
         repo = eitem['origin']
         return repo
 
+    def get_time_to_first_attention(self, item):
+        """Get the first date at which a comment or reaction was made to the issue by someone
+        other than the user who created the issue
+        """
+        comment_dates = [parser.parse(comment['created_at']).replace(tzinfo=None) for comment
+                         in item['comments_data'] if item['user']['login'] != comment['user']['login']]
+        reaction_dates = [parser.parse(reaction['created_at']).replace(tzinfo=None) for reaction
+                          in item['reactions_data'] if item['user']['login'] != reaction['user']['login']]
+        reaction_dates.extend(comment_dates)
+        if reaction_dates:
+            return min(reaction_dates)
+        else:
+            return None
+
     @metadata
     def get_rich_item(self, item):
         rich_issue = {}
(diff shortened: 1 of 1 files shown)


grimoire_elk/enriched/github.py
```diff
@@ -343,6 +359,12 @@ def get_rich_item(self, item):
         if 'project' in item:
             rich_issue['project'] = item['project']
 
+        if issue['comments'] + issue['reactions']['total_count'] == 0:
```
valeriocos: I would convert the if-then-else block to an if-then one, moving out `rich_issue['time_to_first_reaction'] = None` and changing the condition accordingly

grimoire_elk/enriched/github.py
```diff
@@ -249,6 +250,21 @@ def get_project_repository(self, eitem):
         repo = eitem['origin']
         return repo
 
+    def get_time_to_first_reaction(self, item):
+        """
```
valeriocos: Please join lines 254 and 255 as it is done here: https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/discourse.py#L124

valeriocos (approved):
Thank you @aswanipranjal for the PR. 
LGTM. I added just some minor comments. I involve also @acs in the review

grimoire_elk/enriched/github.py
```diff
@@ -249,6 +250,20 @@ def get_project_repository(self, eitem):
         repo = eitem['origin']
         return repo
 
+    def get_time_to_first_reaction(self, item):
+        """Get the first date at which a comment or reaction was made to the issue by someone
```
acs: I would use better `get_time_to_first_attention`. In general we are using attention and not reaction, so it is better to use the same semantic in all places.

acs (changes requested):

aswanipranjal:
@valeriocos @acs I've made the required changes.

acs (approved):

acs:
Nice wotk @aswanipranjal

PR #384, opened 2018-06-26
[enrich][redmine] Add `assigned_to` fields

acs:
`assigned_to` fields are included after processing the
corresponding identity in SortingHat.


alpgarcia:
LGTM. Thanks!

PR #385, opened 2018-06-28
[sortinghat] Align ELK with changes in sortinghat

valeriocos:
Due to enhancements in sortinghat, this code aligns the code to handle sortinghat AlreadyExist exceptions in ELK.


acs (approved):
LGTM. The ex.id is the identifier included in this error:

'UniqueIdentity \\'3870e4416746d9bec42ef0bf4a8f1bb8592fcde0\\' already exists in the registry'

so it is the old uuid.

PR #387, opened 2018-06-28
[mattermost] Fix tests for arthur params

acs:


acs:
@valcos could you review it? Just to confirm after merge that all is ok.

PR #388, opened 2018-06-28
[enrich] Fix the logic for finding the uuid for a SH identity

acs:
Some changes in SH API has broken how gelk finds the uuid for an identity.
This commit fixes it and use the new low level API from SH making the
logic simpler.


valeriocos (approved):
LGTM

PR #391, opened 2018-07-03
[tests] Add support for test using SSL

alpgarcia:
When using secure envs, for instance those versions
of Kibiter including Search Guard, test will fail
if certificate cannot be verified. Including this
change, certificate validation is disabled so tests
can be run without the need of a valid certificate.


valeriocos (approved):
LGTM

valeriocos:
@acs can we merge it?

acs (approved):
LGTM

Issue #363, opened 2018-06-04
[studies] Studies raise exception when cannot verify SSL certificate


jgbarah:
When using indexes in Elasticsearch accessed via SSL, studies fail if the certificate cannot be verified. In the rest of elk, we have a flag to deal with these cases (`insecure` parameter in `enriched.utils.grimoire_con`). It seems there is nothing similar for studies, and we would need it.

The problem manifests with an exception such as (when called from Mordred):

```
Traceback (most recent call last):
  File "/usr/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.5/dist-packages/mordred/task_manager.py", line 92, in run
    task.execute()
  File "/usr/local/lib/python3.5/dist-packages/mordred/task_enrich.py", line 291, in execute
    raise e
  File "/usr/local/lib/python3.5/dist-packages/mordred/task_enrich.py", line 289, in execute
    self.__studies()
  File "/usr/local/lib/python3.5/dist-packages/mordred/task_enrich.py", line 250, in __studies
    do_studies(enrich_backend, studies_args)
  File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/elk.py", line 479, in do_studies
    raise e
  File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/elk.py", line 476, in do_studies
    study(enrich_backend, **params)
  File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/enriched/git.py", line 772, in enrich_areas_of_code
    exists_index = out_conn.exists()
  File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/enriched/ceres_base.py", line 275, in exists
    return self._es_conn.indices.exists(index=self._es_index)
...
```

valeriocos:
This issue has been fixed with: https://github.com/chaoss/grimoirelab-elk/pull/392

PR #392, opened 2018-07-03
[studies] Fix ES conns to use insecure parameter

alpgarcia:
These changes solve #363

It also fixes constructor method, where `insecure` parameter
was not used. `True` was used by default, ignoring the
parameter (which value was in fact `True` by default)


valeriocos (approved):
LGTM

acs (approved):
LGTM

PR #393, opened 2018-07-04
[raw] Add parser params for perceval mattermost backend

valeriocos:
This code parses the parameters for the mattermost perceval backend, thus enabling the execution of this backend in ELK.


acs (approved):
LGTM

PR #395, opened 2018-07-06
[enrich] Include ocean backend to studies

valeriocos:
This code adds the ocean backend as parameter for studies. This modification is required since some studies need access to the raw index (which could be stored in a different machine).

### DIFF
--- a/grimoire_elk/elk.py
+++ b/grimoire_elk/elk.py
@@ -466,14 +466,14 @@ def get_ocean_backend(backend_cmd, enrich_backend, no_incremental,
     return ocean_backend
 
 
-def do_studies(enrich_backend, studies_args):
+def do_studies(ocean_backend, enrich_backend, studies_args):
     for study in enrich_backend.studies:
         selected_studies = [(s['name'], s['params']) for s in studies_args if s['type'] == study.__name__]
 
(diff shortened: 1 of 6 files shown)


grimoire_elk/elk.py
```diff
@@ -466,14 +466,14 @@ def get_ocean_backend(backend_cmd, enrich_backend, no_incremental,
...
-def do_studies(enrich_backend, studies_args):
+def do_studies(ocean_backend, enrich_backend, studies_args):
     for study in enrich_backend.studies:
         selected_studies = [(s['name'], s['params']) for s in studies_args if s['type'] == study.__name__]
 
         for (name, params) in selected_studies:
             logger.info("Starting study: %s, params %s", name, str(params))
             try:
-                study(enrich_backend, **params)
+                study(ocean_backend, enrich_backend, **params)
```
acs: With this change, we need to modify all studies to support this new param, right?

valeriocos: yes, you are right! demography (aoc and onion) have this param included in this PR. Kip don't (so far), I'm checking whether this param is needed there

acs (approved):
LGTM

PR #401, opened 2018-07-11
[enriched] Add code to calculate github pull requests enriched data

### LINKED ISSUE #394: [github pull requests] Add support for github pull requests
Currently grimoire-elk supports parsing github issues data and is unable to parse pull requests data if fetched using perceval.
This issue is regarding added support to calculate enriched data for github pull requests.

To start, I think we have the following points to consider:-
- [x] We can fetch the pull_request data from perceval using the --category pull_request flag
- [X] We can have a different elasticsearch raw index to store the PR data fetched from perceval and similarly a different elasticsearch enriched index to store the PR enriched data
- [X] We need a github_pr backend for the raw as well as enriched index, in elk

aswanipranjal:
Added code to calculate
- Enriched PR data
- The following GMD code development metrics:
  - forks
  - code merge duration
  - number of review comments
  - Maintainer Response to Merge Request Duration

closes #394

### DIFF
--- a/grimoire_elk/enriched/github.py
+++ b/grimoire_elk/enriched/github.py
@@ -56,6 +56,9 @@ class Mapping(BaseMapping):
         mapping = """
         {
             "properties": {
+               "merge_author_geolocation": {
+                   "type": "geo_point"
+               },
                "assignee_geolocation": {
                    "type": "geo_point"
                },
@@ -76,7 +79,8 @@ class GitHubEnrich(Enrich):
 
     mapping = Mapping
 
-    roles = ['assignee_data', 'user_data']
+    issue_roles = ['assignee_data', 'user_data']
+    pr_roles = ['merged_by_data', 'user_data']
 
     def __init__(self, db_sortinghat=None, db_projects_map=None, json_projects_map=None,
                  db_user='', db_password='', db_host=''):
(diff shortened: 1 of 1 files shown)


aswanipranjal:
@acs, @valeriocos, @jgbarah Please review?

valeriocos (approved):
LGTM. @acs if you agree, we can merge this PR

grimoire_elk/enriched/github.py
```diff
@@ -316,7 +334,117 @@ def enrich_onion(self, ocean_backend, enrich_backend,
...
+        if pull_request['state'] != 'closed':
+            rich_pr['time_open_days'] = \
+                get_time_diff_days(pull_request['created_at'], datetime.utcnow())
+        else:
+            rich_pr['time_open_days'] = rich_pr['time_to_close_days']
+
+        rich_pr['user_login'] = pull_request['user']['login']
+        user = pull_request['user_data']
+
+        if user is not None and user:
```
acs: All the participants data must come from SH directly, We must not add it here unless we need data that is not included in SH.

acs (changes requested):

grimoire_elk/enriched/github.py
acs: No, in both cases we need to generate the identities information from SH, like we do in the rest of enrichers. At least, the information that it is included in SH. Using the git enriched, this is done using:

* Roles to be included: https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/git.py#L87
* Addition of identities information: https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/git.py#L415

I need to take a look to fix how identities are filled in github enricher for issues. If you want to play with this idea and reach a PR; I will be happy to review it.

aswanipranjal: Okay, I'll try to understand better how identities are being added in the github enricher and make a PR.

Can we merge this for now, then? Because I'd also like to add the fields for the GMD code development metrics which are dependent on github pull request enrich data.

acs: Yes, I think so. If it is following what is done in github issues enricher. once we fix the current one, we can fix the new one.

acs (approved):

PR #404, opened 2018-07-12
[raw] Fix mapping Twitter

valeriocos:
This code modifies the mapping of the raw items coming from Twitter, according to the perceval format.


acs (approved):
LGTM

PR #407, opened 2018-07-15
[studies] Use hash instead of plain text as id

alpgarcia:
The old ids were too long, so this change uses a hash instead, based on the same fields.


acs:
@alpgarcia why do you need to use a hash instead of the plain text? Is if for fixing a bug or just to create more robust unique ids following how perceval does it?

acs:
After talking with @alpgarcia the issue is:

```
2018-07-13 13:04:53,324 - grimoire_elk.elk - ERROR - Problem executing study enrich_areas_of_code, TransportError(400, 'action_request_validation_exception', 'Validation Failed: 1: id is too long, must be no longer than 512 bytes b
ut was: 520;2: id is too long, must be no longer than 512 bytes but was: 514;3: id is too long, must be no longer than 512 bytes but was: 517;4:..........
```

and with the hash id, the issue is gone! So LGTM!

PR #408, opened 2018-07-16
[enriched] Use metadata__update_on to index tweets

valeriocos:
This code uses the perceval attribute `metadata__updated_on` to search raw tweets and enrich them. This change is due to the fact that the attribute `created_at` in tweets is not recognized by ElasticSearch as a valid date, thus, the raw tweets are not enriched and not visualized in dashboards.


acs (approved):
LGTM

acs (approved):
LGTM

PR #410, opened 2018-07-18
[utils] Goodbye twitter scripts

valeriocos:
With the aim of reducing technical debt and improving coherence within the GrimoireLab components, a new Perceval backend for Twitter has been implemented. Thus, older scripts used to fetch tweets from the Twitter Search API can be decommissioned.


acs (approved):
LGTM

PR #415, opened 2018-07-23
[elastic_items] Remove condition for old twitter items

valeriocos:
This code removes the condition to sort twitter items based on the old attribute `@timestamp`.


acs (approved):
LGTM

PR #416, opened 2018-07-24
[all] change shebang line 

valeriocos:
This PR modifies the shebang line for all Python files in ELK, thus now the line is `#!/usr/bin/env python3` instead of `#!/usr/bin/python3`


acs (commented):
LGTM

PR #417, opened 2018-07-24
[aoc] Replace message commit with first 80 chars

valeriocos:
This code replaces the content of commit messages with empty str, in order to limit the size of enriched indexes.

### DIFF
--- a/grimoire_elk/enriched/study_ceres_aoc.py
+++ b/grimoire_elk/enriched/study_ceres_aoc.py
@@ -140,6 +140,8 @@ class AreasOfCode(CeresBase):
     :param self._git_enrich: GitEnrich object to manage SortingHat affiliations.
     """
 
+    MESSAGE_MAX_SIZE = 80
+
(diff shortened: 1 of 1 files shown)


sduenas (commented):
I think this is necessary due to indexes grow fast with repositories which have long message commits (i.e Eclipse).

A better solution would be to replace the message by the summary of the commit (first line of a commit) but right now this might work.

What do you think @acs?

jgbarah:
If it is not very difficult, having the first line (or even the first 80 chars, or up to the first new line, whatever is first) would be much better. This would for allow to have panels or reports with the summary commits, which may be very valuable.

acs:
I think your proposal of conserving the first line is a better option. Maybe @valeriocos could adapt the commit to this idea.

valeriocos:
I have just modified the PR according to @jgbarah suggestion, now only the first 80 chars are kept.

sduenas (approved):
LGTM

PR #418, opened 2018-07-31
[aoc] Fix study to do nothing if no events

alpgarcia:
When no events are extracted from a set commits, i.e.,
there are no file actions, ignore that block and continue
processing the next batch of commits.


sduenas (approved):
LGTM

PR #419, opened 2018-08-05
[studies] Add enrich_pull_requests study

aswanipranjal:
This code adds the `enrich_pull_requests` study to the enriched/github.py
file. This study aims at adding additional fields that are related to
pull request items but are not fetched by perceval directly for items of the
pull_requests category.
This study should be used using the functionality added by the PR:
https://github.com/chaoss/grimoirelab-sirmordred/pull/190.

This study takes in an input param: `raw_issues_index` which should contain
the data fetched from the same repository from which the pull requests data
is fetched. Using this `raw_issues_index`, the additional data is extracted from
related issue and that data is enriched and added to the relevant pull request
and updated in the enriched pull request index.

The specific fields that are added after running the study are:
- `time_to_merge_request_response`: This was already present in the pull requests enriched index. This field is updated to show the minimum date between a first review or a first comment
- `num_comments`: number of `comments` (not review comments) made on the pull request
- `pr_comment_duration`: the difference between the creation date of the pull request and the latest comment on the pull request.
- `pr_comment_diversity`: the number of different people discussing (commenting) on the pull request

for the last 2, should we consider the review comments also?

### DIFF
--- a/grimoire_elk/enriched/github.py
+++ b/grimoire_elk/enriched/github.py
@@ -25,9 +25,12 @@
 import json
 import logging
 import re
+import time
 
+import requests
 from datetime import datetime
-from dateutil import parser
+
+from grimoirelab_toolkit.datetime import str_to_datetime
 
 from .utils import get_time_diff_days
 
@@ -88,6 +91,7 @@ class GitHubEnrich(Enrich):
 
         self.studies = []
         self.studies.append(self.enrich_onion)
+        self.studies.append(self.enrich_pull_requests)
 
         self.users = {}  # cache users
         self.location = {}  # cache users location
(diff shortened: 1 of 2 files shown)


aswanipranjal:
@acs @jgbarah @valeriocos please review?

grimoire_elk/enriched/github.py
```diff
@@ -281,6 +284,21 @@ def get_time_to_merge_request_response(self, item):
             return min(review_dates)
         return None
 
+    def get_latest_comment_date(self, item):
+        """Get the date of the latest comment on the issue/pr"""
+
+        comment_dates = [parser.parse(comment['created_at']).replace(tzinfo=None) for comment
```
valeriocos: I would use grimoirelab-toolkit, which contains a function that performs the same operation (https://github.com/chaoss/grimoirelab-toolkit/blob/master/grimoirelab_toolkit/datetime.py#L97)

valeriocos:
LGTM, @aswanipranjal would be possible to add some tests or it is too complex?

aswanipranjal:
@valeriocos I can add some tests for this by next week. Right now I am focusing on the producing the manuscripts reports. Will that be okay?

aswanipranjal:
Thank you @valeriocos! @jgbarah Ping. Task#5 is finally complete! (Once https://github.com/chaoss/grimoirelab-sirmordred/pull/190 gets merged, that is)

PR #420, opened 2018-08-09
[enrich-slack] Use `num_members` attribute to count channel members

valeriocos:
This code uses the attribute `num_members` in a Perceval Slack JSON doc to count the members of a Slack channel. However, in order to avoid breaking changes, the former variable `member` is kept and used only if `num_members` is not present in the input document.


sduenas (approved):
LGTM

PR #423, opened 2018-08-23
[tests] Enable tests from no-localhost ES

valeriocos:
This code allows to run the tests against an ES which is not in a local machine. This modification applies also to ESs running in a docker image.


sduenas (approved):
LGTM

PR #424, opened 2018-08-23
[enrich-git] Handle commits with missing user info

valeriocos:
This code allows to handle commits for which no author/committer are defined. For these (rare) cases, the name of the user is set to `Unknown`, while the email to None.


sduenas (approved):
LGTM

PR #425, opened 2018-08-23
[enrich-jira] Improve jira enricher

valeriocos:
This code improves the Jira enricher by:
- replacing the term `assignee_to` with `assignee` (the former doesn't appear to be valid, since it's not defined in the corresponding CSV file).
- aligning the logic to extract identities to the one in the github enricher.


sduenas (approved):
LGTM

PR #427, opened 2018-08-27
[enriched-jira] Set Unassigned to null assignee orgs and usernames

valeriocos:
This code sets all `assignee_name`, `assignee_org_name` and`assignee_user_name` null attributes of issues to `Unassigned`, thus avoiding to visualize results with no values.


alpgarcia:
LGTM, thanks!

PR #428, opened 2018-08-29
[enrich] Set Unknown to null user data (org, name and username)

valeriocos:
This PR sets the value `Unknown` to `*_name`, `*_org_name` and `*_user_name` attributes for each role in an enriched item having null information. Furthermore, it removes the logic to detect Unknown users in JIra and Git enrichers.


grimoire_elk/enriched/enrich.py
```diff
@@ -69,6 +69,8 @@
 DEFAULT_DB_USER = 'root'
 CUSTOM_META_PREFIX = 'cm'
 
+SH_ATTRIBUTE_DEFAULT_VALUE = 'Unknown'
```
acs: SH_ATTRIBUTE_DEFAULT_VALUE is not so clear. I would use SH_UNKNOWN_VALUE

acs (changes requested):
Small change requested an LGTM this PR

PR #429, opened 2018-08-30
[enriched] Update mediawiki project json params

alpgarcia:
This change allows to add a second URL to `projects.json`
file for mediawiki.

In that case, the first URL will be used for retrieving raw data
using Perceval, while the second one will be used to build
the URL of the wiki page. If both, API and web server, uses
the same URL, it is not needed to write the same URL twice,
just let one URL as parameter instead of two.


grimoire_elk/raw/mediawiki.py
```diff
@@ -63,3 +63,21 @@ class MediaWikiOcean(ElasticOcean):
...
+
+        return params
+
+    @classmethod
+    def get_arthur_params_from_url(cls, urls):
+        # The URLs are  the API URL and the wiki URL
+        # We just need the API URL
+        data = urls.split()
+
+        return {"uri": data[0], "url": data[0]}
```
valeriocos: I guess this should be just `return {'"url": data[0]}`, since [the perceval backend](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/mediawiki.py#L502) requires only the `url`.

Have you tested this code with Arthur?

alpgarcia:
Changes added.

This change needs to be supported in mordred too: https://github.com/chaoss/grimoirelab-sirmordred/pull/207

valeriocos:
LGTM, thanks @alpgarcia

valeriocos:
@acs, if you agree with the changes, I would merge this PR

acs (approved):
LGTM

PR #430, opened 2018-08-30
[enrich-meetup] Replace lists with yields

valeriocos:
This code optimizes the memory consumption of the meetup enricher. Thus, it replaces lists to yield statements.


acs (approved):
LGTM

PR #431, opened 2018-08-30
[elk] Modify logic to load new identities to SH

valeriocos:
This code modifies the logic of the method `load_identities` with the intent of improving the memory consumption of ELK. New identities are loaded in bulks of 100, limiting the size of the list containing them.


acs (approved):
LGTM

PR #432, opened 2018-08-30
[enrich-mbox] Replace lists with yields

valeriocos:
This code optimizes the memory consumption of the mbox enricher. Thus, it replaces the list in the method `get_identities` with a yield statement.


acs (approved):
LGTM

PR #433, opened 2018-09-04
[enrich-meetup] Fix error printing the number of comments and RSVPs

sduenas:
This error was introduced in bb91ac23bc8644706116c0637b920ebd4ecd66e6 when lists were replaced by iterators. `ncomments` and `nrsvps` variables are integers and not lists.


valeriocos:
LGTM

PR #439, opened 2018-09-14
[enrich-stackexchange] Add 'origin' and 'tag' fields to answer items

### LINKED ISSUE #436: Wrong project value for Stack Exchange answers
The project value is set to the default value (in this case `Main`) even when the repository is defined in the list of projects. This is caused because enriched Stack Exchange answers do not contain the field `origin`.

sduenas:
So far, Stack Exchange enriched answers did not contain these two attributes. This caused some errors, like to assing a project to an answers due to the mapping between project and answers was done using 'origin' field.

Fixes #436.


valeriocos (approved):

valeriocos:
LGTM

PR #440, opened 2018-09-18
[enrich-mediawiki] Fix unique id for enriched items

valeriocos:
By default, the unique id of an item is the `uuid` that comes from Perceval which is also used to update items during autorefresh operations. In the case of mediawiki, the unique id of an item
relies on the revision id (`revision_revid`). Thus this code removes the method `get_field_unique_id_review` and uses its content to overwrite the `get_field_unique_id` method, avoding in this way misalignments between the unique id used for enrichment and autorefresh operations.


acs (approved):
LGTM

PR #441, opened 2018-09-19
[enrich] Set identities fields to UNDEFINED for null values

sduenas:
When there are identities that are not found in SortingHat, the fields in the enriched items related to those identities are set to an empty string (''). When someone sees this value, it is not clear either there is an error in the visualization, in the way this value was obtained or if the value was blank in the origin.

For these reasons, the new default value will be "-- UNDEFINED --", so these kind of problems can be tracked more easy.

### DIFF
--- a/grimoire_elk/enriched/enrich.py
+++ b/grimoire_elk/enriched/enrich.py
@@ -626,10 +626,10 @@ class Enrich(ElasticItems):
                     break
         return enroll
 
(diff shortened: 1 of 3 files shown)


valeriocos:
LGTM, should I merge it @sduenas ?

valeriocos:
Thanks!

PR #442, opened 2018-09-26
[demography] Catch conn exceptions when fetching author commits

sduenas:
Patch to add a try..except block to catch connection exceptions when getting commits during the demography study.


valeriocos:
Is it possible to add some test code for this PR?

valeriocos:
After discussion, it seems that testing that code is somehow complicated. Thus, I proceed to merge this PR

valeriocos (approved):
LGTM

PR #444, opened 2018-10-02
[enrich-stackexchange] Handle KeyError exception

valeriocos:
This code allows to prevent keyerror exceptions over a question having `answer_count` greater than 0 but not list of `answers` associated.

### DIFF
--- a/grimoire_elk/enriched/stackexchange.py
+++ b/grimoire_elk/enriched/stackexchange.py
@@ -153,7 +153,9 @@ class StackExchangeEnrich(Enrich):
             eitem['question_has_accepted_answer'] = 0
             eitem['question_accepted_answer_id'] = None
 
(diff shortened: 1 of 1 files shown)


grimoire_elk/enriched/stackexchange.py
```diff
@@ -153,7 +153,10 @@ def get_rich_item(self, item, kind='question', question_tags=None):
             eitem['question_has_accepted_answer'] = 0
             eitem['question_accepted_answer_id'] = None
 
-            if question['answer_count'] >= 1:
+            if question['answer_count'] >= 1 and 'answers' not in question:
```
sduenas: Does not work and `elif` block better?

sduenas (approved):
LGTM

PR #445, opened 2018-10-04
[gitlab] Add milestone attribute to enriched items

valeriocos:
This code adds milestone information to enriched items coming from gitlab issues and merge requests. Milestone data is stored in the attribute `milestone`, in case no milestone is present, the value is set to `no-milestone`. Tests have been modified accordingly.

### DIFF
--- a/grimoire_elk/enriched/gitlab.py
+++ b/grimoire_elk/enriched/gitlab.py
@@ -38,6 +38,7 @@ from ..elastic_mapping import Mapping as BaseMapping
 logger = logging.getLogger(__name__)
 
 GITLAB = 'https://gitlab.com/'
+NO_MILESTONE_TAG = "-empty-"
 
 
 class Mapping(BaseMapping):
@@ -233,6 +234,8 @@ class GitLabEnrich(Enrich):
         rich_issue['gitlab_repo'] = re.sub('.git$', '', rich_issue['gitlab_repo'])
         rich_issue["url_id"] = issue['web_url'].replace(GITLAB, '')
 
+        rich_issue['milestone'] = issue['milestone']['title'] if issue['milestone'] else NO_MILESTONE_TAG
+
         if self.prjs_map:
             rich_issue.update(self.get_item_project(rich_issue))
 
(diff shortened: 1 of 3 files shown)


grimoire_elk/enriched/gitlab.py
```diff
@@ -38,6 +38,7 @@
 logger = logging.getLogger(__name__)
 
 GITLAB = 'https://gitlab.com/'
+NO_MILESTON_TAG = "no-milestone"
```
sduenas: Not sure if this is the right value for these cases. What about something like `-empty-`, `-none-` or
`-unassigned-`? We should have a common criteria for these kind of cases.

tests/test_gitlab.py
```diff
@@ -26,6 +26,7 @@
 
 from base import TestBaseBackend
 from grimoire_elk.raw.gitlab import GitLabOcean
+from grimoire_elk.enriched.gitlab import NO_MILESTON_TAG
 
 
 class TestGit(TestBaseBackend):
```
sduenas: Git test?

sduenas (changes requested):
Just minor comments about the code. Please check them.

grimoire_elk/enriched/gitlab.py
valeriocos: I'm not sure we should have a common value. Feel free to propose the no_milestone_tag. I'll update the PR with the new value

tests/test_gitlab.py
valeriocos: It was there before and I didn't modify it. Anyway, I have just refreshed the PR to fix the typo

grimoire_elk/enriched/gitlab.py
valeriocos: done!

grimoire_elk/enriched/gitlab.py
```diff
@@ -38,6 +38,7 @@
 logger = logging.getLogger(__name__)
 
 GITLAB = 'https://gitlab.com/'
+NO_MILESTON_TAG = "-empty-"
```
sduenas: There's a typo. It should be MILESTONE.

sduenas (changes requested):
Minor typo you have to fix before merging this patch.

grimoire_elk/enriched/gitlab.py
valeriocos: done

sduenas (approved):

PR #446, opened 2018-10-04
[gitlab] Add csv for gitlab merges and issues

valeriocos:
This code includes the CSV files for gitlab merges and issues. The description has been taken from the current indexes deployed in production, plus the file `milestone` which is going to be added soon.


schema/gitlab_merges.csv
```diff
@@ -0,0 +1,47 @@
...
+code_merge_duration,float
+created_at,date
+gitlab_repo,keyword
+grimoire_creation_date,date
+id,long
+id_in_repo,keyword
+is_gitlab_merge_request,long
+item_type,keyword
+labels,keyword
+merge_author_login,keyword
```
sduenas: I miss `merge_author_domain`,`merge_author_location`, `merge_author_org`.

sduenas (changes requested):
I miss some fields on the CSV.

valeriocos:
We discussed that the right format for CSVs is the one available for git (https://github.com/chaoss/grimoirelab-elk/blob/master/schema/git.csv), but your comments seem to overlook it. Should I address your comments or modify the full CSV to make it conform to the correct format?

sduenas:
> Should I address your comments or modify the full CSV to make it conform to the correct format?

Yes, you should. Let's add those fields and we can merge this PR.

sduenas (approved):
LGTM

PR #448, opened 2018-10-05
[aoc] Add a constant for out_index default value 

alpgarcia:
Out index name is now accesible as a constant within Git class.


valeriocos (approved):
LGTM

PR #449, opened 2018-10-09
 [onion] Add refresh interval

alpgarcia:
A new parameter `seconds` can be used to define the minimum
time from one study execution to the following. By default has
been set to one week, but can be set for each data source in
`setup.cfg`. Onion will only run again if at least that number
of seconds has passed from the enrichment of the most recent
item in the index.

For information purposes, each time the study is not run due
to this parameter, a log line is printed containing that
minimum date for reference.

Note that the study won't be executed exactly at that time,
but at some point after that time. Studies are executed
after each enrichment processes.

A second commit has been included to solve a bug that prevented the alias to be created after executing the study in some cases:

```
[onion] Fix `all_onion` alias creation  …
Alias must be created if it doesn't exist for the
corresponding `out_index`. There was a bug when checking it
because code was looking only for alias to exist. As
`all_onion` points to different indices (one per data source)
it could exist pointing to another source, which was a false
positive that prevented the alias to be created.

This commit allows to check if the alias exists pointing to
the particular index we are interested in.
```

### DIFF
--- a/grimoire_elk/enriched/ceres_base.py
+++ b/grimoire_elk/enriched/ceres_base.py
@@ -284,12 +284,12 @@ class ESConnector(Connector):
 
         return self._es_conn.indices.put_alias(index=self._es_index, name=alias_name)
 
-    def exists_alias(self, alias_name):
+    def exists_alias(self, alias_name, index_name=None):
         """Check whether or not the given alias exists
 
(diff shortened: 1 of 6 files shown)


sduenas (approved):
LGTM. It's been tested and it can be merged too.

PR #450, opened 2018-10-11
[raw-jira] Select jira project from projects.json url

valeriocos:
This code allows to read the project param of the Jira Perceval backend from the corresponding projects.json url. Thus, it enables to collect and enrich only issues of specific projects within the Jira server.

In order to use this feature, the url within the projects.json should be defined in the following way (ODP is the name of the Jira project):
```
"jira": [
           "http://finosfoundation.atlassian.net ODP"
        ], ..
```

### DIFF
--- a/grimoire_elk/raw/jira.py
+++ b/grimoire_elk/raw/jira.py
@@ -87,7 +87,29 @@ class JiraOcean(ElasticOcean):
     @classmethod
     def get_arthur_params_from_url(cls, url):
         """ Get the arthur params given a URL for the data source """
-        return {"url": url}
+        params = {}
+
+        tokens = url.split(' ')
+        params['url'] = tokens[0]
+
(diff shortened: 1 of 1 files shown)


valeriocos:
The information about the project that includes the issue is already contained in the corresponding enriched item (i.e., attributes `project_id`, `project_key` and `project_name`). Thus, the filtering for a given project can be done directly in the Jira panel (without modifying the enricher).

This PR could be modifed to enable the enricher to create an index with only the information of a given project. What do you think @sduenas , I modify the PR or close it?

valeriocos:
do we close this PR @sduenas ?

valeriocos:
With the recent changes in ELK, this PR is going to be closed and replaced by https://github.com/chaoss/grimoirelab-elk/pull/563

PR #453, opened 2018-10-16
[enrich-bugzilla] Replace list with yield

valeriocos:
This code replaces the list used in the method `get_identities` with a yield, thus reducing the memory footprint.


sduenas (approved):
LGTM

PR #462, opened 2018-10-18
[elastic_items] Refactor `fetch` method

valeriocos:
This code refactors the `fetch` method, removing the infinite loop and break statements.


sduenas (approved):
LGTM

PR #463, opened 2018-10-18
[elk] Add condition to skip loading identites

valeriocos:
This PR prevents to execute the load identites operation for those enrichers that don't return items with identities data. Thus, the proposed code improves the performance of the platform.


sduenas (approved):
LGTM

sduenas:
@valeriocos there are some conflicts with this, can you resolve them and rebase it?

valeriocos:
done @sduenas !

PR #484, opened 2018-10-23
[enrich-gitlab] Handle missing milestone attribute in raw items

valeriocos:
This code handles old raw items coming from the GitLab API that don't contain the milestone attribute.


sduenas (approved):
LGTM

PR #485, opened 2018-10-25
[ELK] Fix flake8 errors

valeriocos:
This code fixes flake8 errors in tests, utils, raw and enriched modules, which appeared after the latest release of the same package.


sduenas (approved):
LGTM

PR #487, opened 2018-10-25
[utils] Goodbye rest script

valeriocos:
This code removes from the utils folder the script `rest.py`, which is not used anymore.


valeriocos:
@acs do you think we can remove the utils/rest.py script? I have checked the code base and it doesn't seem to be used anywhere.

acs (approved):
LGTM

PR #489, opened 2018-10-30
[utils] Avoid enriching items over and over

valeriocos:
This code prevents to trigger enrichment operations over all raw items for backends not having neither from_date or offset attributes in their signatures. This is the case of backends like `dockerhub`, `googlehits`, `jenkins`, `rss` and `twitter`.

### DIFF
--- a/grimoire_elk/elk.py
+++ b/grimoire_elk/elk.py
@@ -455,7 +455,10 @@ def get_ocean_backend(backend_cmd, enrich_backend, no_incremental,
         elif 'offset' in signature.parameters:
             ocean_backend = connector[1](backend, offset=last_enrich)
         else:
-            ocean_backend = connector[1](backend)
+            if last_enrich:
+                ocean_backend = connector[1](backend, from_date=last_enrich)
+            else:
+                ocean_backend = connector[1](backend)
     else:
         # We can have params for non perceval backends also
         params = enrich_backend.backend_params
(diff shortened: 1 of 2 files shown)


valeriocos:
@acs could you have a look at this PR? My feeling is that the code could be simplified even more, like this. What do you think?
```
def get_last_enrich(backend_cmd, enrich_backend):
    if backend_cmd:
        backend = backend_cmd.backend
        # Only supported in data retrieved from a perceval backend
        # Always filter by repository to support multi repository indexes
        backend_name = enrich_backend.get_connector_name()
        filter_ = get_repository_filter(backend, backend_name)
        last_enrich = enrich_backend.get_last_update_from_es([filter_])
    else:
        last_enrich = enrich_backend.get_last_update_from_es()

    return last_enrich
```

valeriocos:
ping @acs

acs (approved):
LGTM

alpgarcia:
It seems enrichment is still not incremental. `last_enrich` is not used at all in:
https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/elk.py#L449

valeriocos:
ok, I'm going to check it

sduenas (changes requested):
We have seen that some code is also needed here:
https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/elk.py#L449

As you can see, there's a doble check about the usage of `from-date`, `offset` or nothing.

alpgarcia:
LGTM, thanks!

alpgarcia:
Merged in e1e49a8809187321bf8e540fc2dfed8036d6d611

PR #490, opened 2018-10-31
[enrich-discourse] Refactor logic to enrich items

valeriocos:
This code changes the logic to enrich items. Now, raw items are enriched on the fly together with enriched comments. Tests have been modified accordingly.

### DIFF
--- a/grimoire_elk/enriched/discourse.py
+++ b/grimoire_elk/enriched/discourse.py
@@ -27,6 +27,9 @@ from .utils import get_time_diff_days, grimoire_con
 
 from .enrich import Enrich, metadata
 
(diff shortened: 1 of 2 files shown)


alpgarcia:
LGTM, thanks!

I did the following tests:
* Uploading just one question with aswers.
* Uploading a set of questions, being the newest the one already processed (to see if that one is updated looking at `metadata__enriched_on`, it must due to the incremental update based on `gte` operator over `metadata__timestamp`)
* Uploading the complete set and checking we have all questions.
* Uploading the same set, but breaking the execution a couple of times and check the result is the same as the previous test.

PR #491, opened 2018-10-31
[enrich-askbot] Refactor logic to enrich items

valeriocos:
This code changes the logic to enrich items. Now, raw items are enriched on the fly together with enriched answers and comments.


sduenas (approved):
LGTM

PR #495, opened 2018-10-31
[elk] Avoid processing None identities

valeriocos:
This code prevents to process empty generators or none objects.


valeriocos:
The idea behind this PR is that since all `get_identities` methods return now a generator, a check should be needed to ensure that the generator is not null (this is the case when no identities are found in a given item).

sduenas (approved):

PR #501, opened 2018-11-07
[enrich-stackexchange] Reorganize tags fields

valeriocos:
This code reorganizes the tags fields for questions and answers.
+ Enriched answers include the fields:
  - `answer_tags` (tags of the answer)
  - `question_tags` (tags of the question)
+ Enrich questions include the fields:
  - `question_tags` (tags of the question)
  - `answers_tags` (all tags of answers)
  - `thread_tags` (the sum of question and answers tags)

### DIFF
--- a/schema/stackoverflow.csv
+++ b/schema/stackoverflow.csv
@@ -3,6 +3,7 @@ answer_status,keyword
 answer_count,long
 answer_id,long
 answer_tags,keyword
+answers_tags, keyword
 author_bot,boolean
 author_id,keyword
 author,keyword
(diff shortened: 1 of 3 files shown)


schema/stackoverflow.csv
```diff
@@ -3,6 +3,7 @@ answer_status,keyword
 answer_count,long
 answer_id,long
 answer_tags,keyword
+answers_tags, keyboard
```
alpgarcia: Keyboard?? :P

alpgarcia (changes requested):
LGTM, after testing it with micro mordred, just a minor change.

schema/stackoverflow.csv
valeriocos: duplicated?

PR #502, opened 2018-11-07
[tests] Check value has_identities of enrichers

valeriocos:
In order to avoid introducing bugs related to loading identities (as it has recently happened for mediawiki), this PR proposes to include a test to check the value of the `has_identities` method for each enricher.


dlumbrer (approved):
LGTM, tested and works!

screenshot from 2018-11-08 13-33-58

PR #503, opened 2018-11-08
[raw-gitlab] Allow passing blacklisted items from projects.json

valeriocos:
This code allows to pass a list of blacklisted item ids (issues or pull requests) via the projects.json.


dlumbrer (approved):
LGTM, thanks @valeriocos !

PR #504, opened 2018-11-09
[grimoire_elk] Goodbye track_items.py

valeriocos:
This code removes the track_items.py which has been replaced by the `filter-raw` feature.


valeriocos:
@acs , it seems that this script is not used anymore. Do you remember what it was used for?
Can we delete it?

valeriocos:
ping @acs

valeriocos:
@sanacl @jgb do you know if this script is still useful?

sduenas:
If nobody says other thing we can remove it

sduenas (approved):
LGTM

PR #505, opened 2018-11-12
[elastic_items] Handle none page

valeriocos:
This code allows to handle none page value returned by `get_elastic_items`.


dlumbrer (approved):
Tested and OK, LGTM, thanks @valeriocos :+1:

PR #506, opened 2018-11-12
[proposal] refactor elk

valeriocos:
The goal of this PR is to propose a refactor of ELK. The commits included in the PR promotes the use of the elasticsearch library instead of simple HTTP requests, thus delegating the responsability of retrieving data to this library.

This PR could be extended in different ways, for instance:
- reorganize the dependencies of ELK, for instance removing the dependencies to arthur and perceval and move the corresponding logic to mordred (which together with arthur take care of scheduling tasks)
- refactor the logic of the enrichers, which currently mix both business and visualization data. The latter data could be moved to another package (e.g., visualization) which prepares the data to be plotted in the dashboards.
- remove code not used anymore. ELK has been continuosly evolving, however some logic is no longer used and could be removed (e.g., utils folder)


sduenas:
I'm going to close it. Probably we'll have to go back to this PR in a while.

PR #507, opened 2018-11-12
[enrich-github] Fix attribute repository

valeriocos:
This code fixes the value of the attribute repository, which is now derived from the origin. This modification is needed for GitHub repos that changed their names and for which the corresponding attribute `html_url` was not updated.


alpgarcia:
LGTM, thanks!

PR #509, opened 2018-11-13
[elk] Refactor `refresh_identities`

valeriocos:
This code refactors the method `refresh_identities` to be able to work with the recent changes in the sortinghat api. In a nutshell, in order to reduce the memory footprint, the sortinghat api
returns a generator of identities (instead of lists), the `refresh_identities` method manipulates this list to prepare a filter used to update the data with the identities information updated.


sduenas (approved):
Reviewed and tested. It works fine.

PR #510, opened 2018-11-13
[elk] Handle aliases

valeriocos:
This PR enhances ELK to create ad-hoc aliases for raw and enriched indexes. The aliases accepted are simple ones (names), filtered aliases are currently not supported.


sduenas (approved):
Reviewed and tested.

PR #513, opened 2018-11-13
[gitlab-raw] Fix mappings to avoid immense term exception

valeriocos:
This code modifies the mappings of raw data to avoid ES insertion
exceptions.


sduenas (approved):
LGTM

PR #515, opened 2018-11-16
[enrich] Prevent find identities for none ids

valeriocos:
This code prevents to query SortingHat when the ids returned by the utils.uuid is none.


grimoire_elk/enriched/enrich.py
```diff
@@ -862,6 +862,10 @@ def __get_sh_ids_cache(self, identity_tuple, backend_name):
             # Find the uuid for a given id.
             id = utils.uuid(backend_name, email=iden['email'],
                             name=iden['name'], username=iden['username'])
+
+            if not id:
```
sduenas: I think this should at least write a warning message.

sduenas (changes requested):

valeriocos:
PR updated!

sduenas (approved):
LGTM

PR #517, opened 2018-11-16
[ELK] Remove credentials from log messages

valeriocos:
This PR allows to remove credentials (username and password) from the log messages which track the URLs that interact with the ElasticSearch server. Thus, it avoids possible security leaks when copy/paste traceback errors in no-safe environments (GitHub, GitLab).


sduenas (approved):
LGTM

PR #518, opened 2018-11-16
[enrich] Use grimoirelab-toolkit methods

valeriocos:
This code replaces the use of dateutil parser with the methods provided by the grimoirelab-toolkit. Thus, promoting homogeneity in the code and reducing the risk of failures.


sduenas (approved):
LGTM

PR #521, opened 2018-11-16
[studies] Improve log lines prefix (rebased)

alpgarcia:
Log lines now include more info about the specific study,
data source or index in order to ease finding them among
other log lines.


alpgarcia:
...and now we have here a W503. I guess having W503 and W504 together doesn't make too much sense here, does it?

sduenas (approved):
It works ok. Thanks!

PR #524, opened 2018-11-21
[elastic] Prevent add alias if already in use in ES

valeriocos:
This code avoids to automatically assign an alias to an index, in case the alias is already used by another index. Thus, operators can manually assign aliases to indexes after having performed quality checks.


sduenas (commented):
LGTM but it must be tested

valeriocos:
@sanacl do you want to test it?

PR #525, opened 2018-11-26
[enrich-gitlab] Fix GitLab enricher

valeriocos:
This PR updates the way the attribute `time_to_close_days` is calculated. In GitLab a merge request can end up in two states, merged and closed. The former concerns merge requests that were finally merged to the code base, while the latter represents rejected merge requests. Given the aforementioned workflow, the code has been changed accordingly. Furthermore, this PR fixes also the calculation of the attribute `merged`, which before was always set to true.

### DIFF
--- a/grimoire_elk/enriched/gitlab.py
+++ b/grimoire_elk/enriched/gitlab.py
@@ -264,10 +264,12 @@ class GitLabEnrich(Enrich):
         # The real data
         merge_request = item['data']
 
-        rich_mr['time_to_close_days'] = \
-            get_time_diff_days(merge_request['created_at'], merge_request['closed_at'])
+        time_to_close_days = get_time_diff_days(merge_request['created_at'], merge_request['closed_at'])
+        time_to_merge_days = get_time_diff_days(merge_request['created_at'], merge_request['merged_at'])
 
-        if merge_request['state'] != 'closed':
+        rich_mr['time_to_close_days'] = time_to_merge_days if time_to_merge_days else time_to_close_days
+
+        if merge_request['state'] not in ['merged', 'closed']:
             rich_mr['time_open_days'] = \
                 get_time_diff_days(merge_request['created_at'], datetime.utcnow())
         else:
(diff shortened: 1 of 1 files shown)


grimoire_elk/enriched/gitlab.py
```diff
@@ -264,10 +264,12 @@ def __get_rich_merge(self, item):
         # The real data
         merge_request = item['data']
 
-        rich_mr['time_to_close_days'] = \
-            get_time_diff_days(merge_request['created_at'], merge_request['closed_at'])
+        time_to_close_days = get_time_diff_days(merge_request['created_at'], merge_request['closed_at'])
```
sduenas: Could you add a comment here explaining what differences are between these two fields and why they are calculated this way? My point is you should explain here what you talked to GitLab folks. I don't want to lose that knowledge.

grimoire_elk/enriched/gitlab.py
```diff
@@ -319,7 +321,7 @@ def __get_rich_merge(self, item):
         rich_mr['created_at'] = merge_request['created_at']
         rich_mr['updated_at'] = merge_request['updated_at']
         rich_mr['url'] = merge_request['web_url']
-        rich_mr['merged'] = True if rich_mr['state'] else False
+        rich_mr['merged'] = True if rich_mr['state'] == 'merged' else False
```
sduenas: This should be done this way:

```suggestion
        rich_mr['merged'] = rich_mr['state'] == 'merged'
```

sduenas (changes requested):
Please address my comments.

grimoire_elk/enriched/gitlab.py
valeriocos: Well the information is in the commit message, however now it's also in the code

grimoire_elk/enriched/gitlab.py
valeriocos: I believe `True if rich_mr['state'] == 'merged' else False` was clearer, however now the code has been changed according to your comment

sduenas: This case is an anti-pattern (https://ucsb-cs24.github.io/topics/antipatterns/)

sduenas (approved):

PR #528, opened 2018-12-07
[enrich-gitlab] Add time fields info

valeriocos:
This PR adds time information to the enriched MRs of GitLab. In particular, the new fields are:
- merged_at: when a MR was merged (it is null if the MR was rejected)
- closed_at: when a MR was closed (it is null if the MR was accepted)
- solved_at: when a MR was solved (merged or closed).

GitLab merge request schema has been updated accordingly.


sduenas (approved):
LGTM

valeriocos:
Sure, it is possible. However the value of `solved_at` for GitHub PR should be equal to `closed_at`.
The workflow of a pull request in GitHub always end to a unique state (`closed`) either if it has been merged or rejected. Thus the value `closed_at` is always set, and you know if a PR has been accepted depending on the value of the boolean attribute `merged`. In GitLab, MRs can end to two states: `closed` (for rejected MRs) and `merged` (for merged MRs), where `closed_at` is set for the former and `merged_at` for the latter.

sduenas:
Understood. Not sure then whether we need it in GitHub PRs.

PR #530, opened 2018-12-21
[ELK] Use the cfg section name to retrieve project name

valeriocos:
This PR replaces the use of the connector name with the cfg section name to retrieve the project name associated to a data source. 
In a nutshell, initially the connector name had a one-to-one correspondence with the cfg section name, which was used to get the project name in the projects.json. Recently, the cfg section name has been enhanced to include also an identifier (e.g., [gitlab:issue]), thus the connector name cannot be used anymore to find the project name and must be replaced with the cfg section name (which is now propagate from mordred to ELK).

Note that in case the cfg section name is null (this is the case when executing the tests in ELK), the connector name is used.

### DIFF
--- a/grimoire_elk/elastic_items.py
+++ b/grimoire_elk/elastic_items.py
@@ -53,6 +53,7 @@ class ElasticItems():
         self.requests = grimoire_con(insecure)
         self.elastic = None
         self.elastic_url = None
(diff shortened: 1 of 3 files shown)


sduenas (approved):
LGTM

PR #533, opened 2018-12-31
[elastic] Change log message level when alias already exists

valeriocos:
This code changes the log message level from warning to debug when an alias already exists. This change aims at reducing the size of logs produced by the platform. The corresponding test has been changed accordingly. 
An additional fix has been provided to ensure the removal of the onion index at the end of the corresponding test. This was needed due to the fact that study aliases cannot be used in more indexes at the same time.

### DIFF
--- a/grimoire_elk/elastic.py
+++ b/grimoire_elk/elastic.py
@@ -240,7 +240,7 @@ class ElasticSearch(object):
         """
         aliases = self.list_aliases()
         if alias in aliases:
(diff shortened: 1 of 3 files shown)


sanacl:
@valeriocos LGTM but checks have failed :face_with_head_bandage:

valeriocos:
thank you @sanacl for the pointer

sanacl:
@valeriocos LGTM.

PR #535, opened 2019-01-02
[ELK] Include stacktrace in log messages

valeriocos:
This code includes the stacktrace information in the log messages generated when exceptions are thrown during the collection, enrichment of the data as well  during the loading of identities.


sduenas (approved):
LGTM

PR #536, opened 2019-01-02
[elastic] Change log message level when inserting data to ES

valeriocos:
This code changes the log message level from INFO to DEBUG when inserting data to ElasticSearch.


sduenas (approved):
LGTM

PR #538, opened 2019-01-18
[ELK] Handle deleted commits

valeriocos:
This PR allows to handle deleted commits in Git raw and enriched indexes. 
A generic method `update_items` is included for raw and enriched connectors in order to update the corresponding indexes just after the collection and enrichment operations. For the case of Git, at the end of the enrichment, the list of current commit SHAs is retrieved from the repository and compared with the hashes stored in the raw index. The different of these two sets gives the deleted commits (the ones that are stored in ES but not present in the repo), which  are finally deleted from the raw and enriched indexes.


sduenas (approved):
LGTM

PR #540, opened 2019-01-30
[elastic] Change log level message when alias is already in use

valeriocos:
This code changes the log level of the message produced when an alias is already in use.


sduenas (approved):
LGTM

PR #541, opened 2019-02-06
Remove alias in use check for onion

alpgarcia:
This commit removes the check for `all_onion` alias that
prevents its creation if it is already in use. This check
only allowed to link one data source to the alias, so onion
data for the rest of the sources remained unlinked.


sduenas (approved):
LGTM

PR #542, opened 2019-02-11
[elk] Use filter-raw to get the last enrich date

valeriocos:
This PR allows to include the `filter-raw` param to identify the last date of enrichment for a given index composed by items having the same origin, but coming from different projects (bugzilla, gerrit, jira). 
In a nutshell, the current way of calculating the last date of enrichment takes into account only the origin and the `metadata_timestamp`, while the proposed solution includes also the `filter-raw` param, thus allowing to identify the last date of enriched items for a given filter (e.g., project).

### DIFF
--- a/grimoire_elk/elastic.py
+++ b/grimoire_elk/elastic.py
@@ -420,10 +420,7 @@ class ElasticSearch(object):
             term = '''{"term" : { "%s" : "%s"}}''' % (filter_['name'], filter_['value'])
             terms.append(term)
 
-        if len(filters_) == 1:
-            data_query = '"query": %s,"' % filters_[0]
-        else:
-            data_query = '''"query": {"bool": {"filter": [%s]}},''' % (','.join(t for t in terms))
+        data_query = '''"query": {"bool": {"filter": [%s]}},''' % (','.join(terms))
 
         data_agg = '''
             "aggs": {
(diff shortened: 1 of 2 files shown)


sduenas (approved):
LGTM

grimoire_elk/elastic.py
```diff
@@ -410,17 +410,20 @@ def get_last_item_field(self, field, filters_=[], offset=False):
...
                 continue
-            data_query += '''
-                "query" : {
-                    "term" : { "%s" : "%s"  }
-                 },
-            ''' % (filter_['name'], filter_['value'])
+            term = '''{"term" : { "%s" : "%s"}}''' % (filter_['name'], filter_['value'])
+            terms.append(term)
+
+        if len(filters_) == 1:
```
sduenas: Why don't use `terms` in both places?

grimoire_elk/elastic.py
```diff
@@ -410,17 +410,20 @@ def get_last_item_field(self, field, filters_=[], offset=False):
...
+            data_query = '"query": %s,"' % filters_[0]
+        else:
+            data_query = '''"query": {"bool": {"filter": [%s]}},''' % (','.join(t for t in terms))
```
sduenas: `terms` is a list so it would work with `','.join(terms)`.

sduenas (changes requested):
Please check my comments and forget my previous review.

sduenas (approved):
LGTM

PR #543, opened 2019-02-12
[elastic_items] Update ES query for incremental enrichment

valeriocos:
This code changes the behaviour of the ES query to retrive items from the raw indexes for incremental enrichment purposes. In a nutshell, the query now uses a QUERY context which returns the items matching the query plus a score that rates how good the match is. However the score is not needed in ELK, thus the content is changed to FILTER context, which doesn't calculate any score.

### DIFF
--- a/grimoire_elk/elastic_items.py
+++ b/grimoire_elk/elastic_items.py
@@ -231,7 +231,7 @@ class ElasticItems():
                 {
                     "query": {
                         "bool": {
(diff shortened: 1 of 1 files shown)


alpgarcia:
LGTM!

sduenas (commented):
I tested it incrementally with Git and it works. I guess it can be merged. Any further comments @alpgarcia?

PR #544, opened 2019-02-12
[studies] Add origin based enrichment in AOC

alpgarcia:
These changes make the study enrich only those repos included
in `projects.json` file instead of enriching the whole set
of data found in the input index.


valeriocos (approved):
LGTM

PR #545, opened 2019-02-15
[ELK] Add GitHub stats support

valeriocos:
This PR provides support to track GitHub repository stats such as the evolution of forks, stars and watchers. Test have been included.

Find below an example about how to use it from mordred:
**setup.cfg**
```
[githubstats]
raw_index = test_github-repo-raw
enriched_index = test_github-repo
api-token = ....
sleep-for-rate = true
no-archive = true
category = repository
```

**projects.json**
```
"githubstats": [
   "https://github.com/chaoss/grimoirelab-perceval"
]
```

### DIFF
--- a/grimoire_elk/enriched/github_stats.py
+++ b/grimoire_elk/enriched/github_stats.py
@@ -1,0 +1,65 @@
+# -*- coding: utf-8 -*-
+#
+#
+# Copyright (C) 2015 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+#
+# Authors:
+#   Valerio Cosentino <valcos@bitergia.com>
+#
+
+import logging
+
+from .enrich import Enrich, metadata
+
+
+logger = logging.getLogger(__name__)
+
+
+class GitHubStatsEnrich(Enrich):
+
+    def get_identities(self, item):
+        """Return the identities from an item"""
+        identities = []
+
+        return identities
+
+    def has_identities(self):
+        """Return whether the enriched items contains identities"""
+
+        return False
+
+    @metadata
+    def get_rich_item(self, item):
+        eitem = {}
+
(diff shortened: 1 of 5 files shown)


grimoire_elk/enriched/github_stats.py
```diff
@@ -0,0 +1,65 @@
+# -*- coding: utf-8 -*-
+#
+#
```
sduenas: Remove this blank line and update the copyright year to include 2019 too.

sduenas: Why is this in a different module? With GitLab categories are managed on the same module.

grimoire_elk/raw/github_stats.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+#
+# StackExchange Ocean feeder
```
sduenas: Same comments as in the enrichment module. Moreover, remove this comment. This is not related to StackExchange.

tests/test_githubstats.py
```diff
@@ -0,0 +1,107 @@
...
+        eitem = enrich_backend.get_rich_item(item)
+        self.assertEqual(eitem['forks'], 1)
+        self.assertEqual(eitem['stars'], 1)
+        self.assertEqual(eitem['watchers'], 1)
+
+    def test_raw_to_enrich_projects(self):
+        """Test enrich with Projects"""
+
+        result = self._test_raw_to_enrich(projects=True)
+        # ... ?
```
sduenas: Why is this comment here and on line 92?

sduenas (changes requested):
Please address my comments.

grimoire_elk/enriched/github_stats.py
valeriocos: GitLab merges and issues share similar mappings (the one that we redefined is common in both), similar thing happens for GitHub issues and pull requests.  Based on the previous implementation of the PR in perceval (https://github.com/chaoss/grimoirelab-perceval/pull/481), the GitHub repo data didn't share things with the issue and pull request mappings, furthermore it didn't include identities information. However, with the current PR in perceval (the repo data includes identities such as `owner`) and if we are fine of having mappings about attributes which don't exist in the data (e.g., `merge_author_geolocation`, `assignee_geolocation`, etc.), this PR can be reworked to add everything in the same module.

valeriocos: > Remove this blank line and update the copyright year to include 2019 too.
If you agree, I would address this comment in another PR, since it seems this applies to most files in ELK.

grimoire_elk/raw/github_stats.py
valeriocos: Perfect, if you agree I would address this comment in another PR (the same as above), since Ocean is an old definition for raw connectors.

tests/test_githubstats.py
valeriocos: When tests were introduced in ELK, these ones were kickstarted with the idea of completing them at some point in the future (same situation in the other tests), but the task was never scheduled.

PR #561, opened 2019-02-21
[enrich] Update Git references data in raw and enriched indexes

valeriocos:
This code updates the information about Git references of the items corresponding to commits in both raw and enriched indexes. The update is needed due to operations (e.g., mergesand branch deletions) which affect the repository references and consequently their commits. Thus, this code aims at ensuring consistency between the information of the original Git repository and the data stored in ElasticSearch.

Note that the enriched Git items have been modified to include the refs info, which was not present there. It may be needed to update the corresponding Git schema and panels, which could be done in another PR.

I performed some performance tests with the grimoirelab-perceval repo:

* 18-19 secs disabling `update_commit_references`
* 38-39 secs enabling `update_commit_references`

### DIFF
--- a/grimoire_elk/enriched/git.py
+++ b/grimoire_elk/enriched/git.py
@@ -318,7 +318,7 @@ class GitEnrich(Enrich):
         self.__fix_field_date(commit, 'CommitDate')
 
         # data fields to copy
-        copy_fields = ["message"]
+        copy_fields = ["message", "refs"]
         for f in copy_fields:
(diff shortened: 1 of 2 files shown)


sduenas:
This PR should be tested with a repository bigger and with more branches than Perceval. 20 seconds to re-enrich this repository means to me too much.

How about to test it with other repositories like:
 * https://github.com/elastic/kibana
 * https://github.com/GNOME/gtk

PR #562, opened 2019-02-22
[raw] Handle filter raw for confluence

valeriocos:
This code allows to parse the filter raw information from the entries within the mordred projects.json.
The projects.json should be changed to the following way:
`"concourse-url --filter-raw=data._expandable.space:/rest/api/space/your-project"`


sduenas (approved):
LGTM

PR #563, opened 2019-02-26
[raw] Handle filter raw for jira

valeriocos:
This PR allows to parse the filter raw information from the entries within the mordred projects.json. The projects.json should be changed to the following way:
`jira-url --filter-raw=data.fields.project.key:TARGET-PROJECT-KEY`

Tests have been added.


sduenas (approved):
LGTM. I also tested it. You can merge this PR.

PR #564, opened 2019-02-27
[enrich] Update branch info in git commits

valeriocos:
This code allows to update the branch information of the JSON documents representing Git commits. First the list of branches is cleared in the enriched index, then the original repo is queried to retrieve the list of commit hashes for each branch and finally the obtained hashes are added to the branches attribute of the documents.

Git schema and enrich/utils have been updated. The former now includes the `branches` attribute, the latter enables the retry mechanism for 409 HTTP errors, which may occur for the operations `update_by_query` and `delete_by_query`.

Find below some performance tests:
- perceval
  - without branch update: 4 secs
  - with branch update: 5 secs
- mordred
  - without branch update: 4 secs
  - with branch update only for enrich: 1min20secs
- elastic/kibana
  - withouth branch update: 2 mins
  - with branch update: 35 mins
  - wich branch update only for enrich: 12 mins

This PR replaces https://github.com/chaoss/grimoirelab-elk/pull/561

### DIFF
--- a/grimoire_elk/enriched/git.py
+++ b/grimoire_elk/enriched/git.py
@@ -30,7 +30,8 @@ import pkg_resources
 import requests
 from elasticsearch import Elasticsearch
 
-from grimoirelab_toolkit.datetime import datetime_to_utc, str_to_datetime
+from grimoirelab_toolkit.datetime import (datetime_to_utc,
+                                          str_to_datetime)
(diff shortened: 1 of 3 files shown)


sduenas (approved):
LGTM.  Maybe it's not ideal regarding the time it needs to update the branches but we can live with it.

PR #565, opened 2019-02-28
[raw] Parse filters passed in the projects.json entries

valeriocos:
This PR redefines the method `get_p2o_params_from_url` to parse the filters declared in the projects.json entries. Filters are identified by the delimiter `--filter`, while the filter name and values are separated by `=`. For instance:
- `https://bugzilla.mozilla.org --filter-raw=data.product:Add-on SDK,data.component:General`
- `http://finosfoundation.atlassian.net --filter-raw=data.fields.project.key:ODP`
- `https://github.com/chaoss/grimoirelab-elk --filter-raw-prefix=data.files.file:README,data.files.file:perceval`

The code provided applies for the Perceval backends that require only an URL as input parameter, for the others, the method `get_p2o_params_from_url` should be redefined. Unit tests have been provided for the new method and live tests have been executed on the following data source:
- [x] Mediawiki
- [x] Discourse
- [x] Jira
- [x] Gerrit
- [x] Bugzilla
- [x] Git

The method `get_p2o_params_from_url` redefined in the raw connectors for git, gerrit, askbot, jira, confluence and bugzilla   have been deleted.

### DIFF
--- a/grimoire_elk/elastic_items.py
+++ b/grimoire_elk/elastic_items.py
@@ -50,7 +50,9 @@ class ElasticItems():
         self.from_date = from_date  # fetch from_date
         self.offset = offset  # fetch from offset
         self.filter_raw = None  # to filter raw items from Ocean
+        self.filter_raw_dict = []
         self.filter_raw_should = None  # to filter raw items from Ocean
+        self.filter_raw_should_dict = []
 
         self.requests = grimoire_con(insecure)
         self.elastic = None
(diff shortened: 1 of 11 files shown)


sduenas:
@valeriocos is this PR ready? I think it needs more work but I'm not sure.

valeriocos:
As reported in the corresponding internal ticket the PR is still not ready. I guess it will be ready today.

valeriocos:
This PR should be reviewed with https://github.com/chaoss/grimoirelab-sirmordred/pull/274

sduenas (approved):
LGTM. 

I have one question, though. Is there anyway to check whether `projects.json` is valid or not when `filter-raw` and `unknown` are in use? I see to use them a little tricky and error prone.
In any case I think the PR can be merged.

valeriocos:
currently there is no support to check the use of filter-raw params. I agree with you that it's error prone and requires some expertise from the operator that set ups the configuration. Maybe some documentation in mordred could help, what you think @sduenas ?

PR #566, opened 2019-03-01
[enrich] Include new fields to Jira and Gerrit enriched items

valeriocos:
This PR enhances the Jira and Gerrit enrichers by including the following attributes to enable the use of painless scripting:
- Jira
 - `updated`: when an issue was updated
- Gerrit
 - `created_on`: when a review was created
- `last_updated`: when a review was updated

The corresponding schemas have been updated accordingly


sduenas (approved):
LGTM

PR #567, opened 2019-03-05
[enriched] Handle 502 and 503 errors

valeriocos:
This PR enables the retry mechanism for 502 and 503 HTTP errors. This change is needed to reduce the connection problems happening in remote instances of ElasticSearch.


sduenas (approved):
LGTM

PR #568, opened 2019-03-05
[raw] Prevent mapping of data.ancestors.extensions for confluence raw…

valeriocos:
This code prevents to index the data included in data.ancestors.extensions in order to avoid parsing_mapping_errors.


sduenas (approved):
LGTM

PR #572, opened 2019-03-07
[elk] Remove filter-raw-* for enrich backend

valeriocos:
This code removes the filter-raw and filter-raw-prefix which were set in the `enrich_backend` function but not used by the enrich backend.


sduenas (approved):
LGTM

PR #573, opened 2019-03-07
[enrich] Add onion study to gerrit

valeriocos:
This code includes the onion study to gerrit. Thus, now an onion study can be defined via the setup.cfg of modred.


sduenas (approved):
LGTM

PR #577, opened 2019-03-12
[elk] Handle project name for filtered repos

valeriocos:
This PR assigns the default project name `Main` to the items coming from the repos included in the `unknown` section within the projects.json. Furthermore, it also removes the method `get_item_project` within the bugzillarest enricher, which logic is not needed anymore (since related to the previous implementation of `--filter-raw`).


sduenas (approved):
LGTM

PR #582, opened 2019-03-19
[elk] Fix incremental enrichment for filter-raw repos

valeriocos:
This PR fixes the incremental enrichment for repos having the filter-raw param. In a nutshell, the last enriched item is calculated in mordred (TaskEnrich) before starting enriching the single repos which data is copied to the same enriched index.
In case of a faulty enrichment execution, the number of raw and enriched items doesn't match, thus the enriched index should be recreated to avoid losing items.


sduenas (approved):
LGTM

PR #583, opened 2019-03-20
[utils] Change assessment of last enrich date

valeriocos:
This code changes the way how the last enrich date is calculated. If the enriched index is empty the last enrich date is set to None. Otherwise, the last enrich date is the minimum between the last metadata_timestamp of the enriched index before starting the TaskEnrich (this covers the case of repos with filter-raw) and the last metadata_timestamp of the items filtered by repo (this covers the case of a list of repos which data is shared in the same index).


sduenas (approved):
LGTM

PR #584, opened 2019-03-20
[tests] Update finosmeetrings backend test to the new schema

sduenas:
Due recent changes in FINOS Meetrings backend, some of the fields of the raw items have been renamed. This commit updates tests data to align them with the current backend.


valeriocos (approved):
LGTM

PR #585, opened 2019-03-21
[elk] Remove identities

valeriocos:
This PR allows to delete unique identities from SortingHat according to a retention time window. The deletion occurs in two ways, the former checks that the deleted unique identities in the enriched indexes are not present in SortingHat, the latter checks that all unique identities in SortingHat are the ones present in the enriched indexes (thus avoiding to store orphan identities which would fall outside the retention time window.)

In a nutshell the proposed code works as follow. For every enrich iteration, the identifiers of the unique identities in the enriched items are retrieved and stored in the index `grimoirelab_identities_cache` together with the time when they were found (`last_seen`). In case the data retention policy is on, the `grimoirelab_identities_cache` index is used to identify the identities not seen before the data retention policy time frame (defined by the param `retention_time`). Such identities are retrieved and deleted from the SortingHat database. 

To detect and remove orphan identities (the ones stored in SortingHat but not present in `grimoirelab_identities_cache`), all unique identites in SortingHat are retrieved and processed in the following way:
- the ones that have all identities in non-active data sources are deleted, while the ones that have some 
  identities in non active data sources are modified in order to remove those identities.
- the unique identities left are compared with the ones in `grimoirelab_identities_cache` and they are deleted if not found.

This PR should be reviewed with https://github.com/chaoss/grimoirelab-sirmordred/pull/283

### DIFF
--- a/grimoire_elk/elastic.py
+++ b/grimoire_elk/elastic.py
@@ -32,7 +32,7 @@ import requests
 from grimoire_elk.errors import ELKError
 from grimoire_elk.enriched.utils import (unixtime_to_datetime,
                                          grimoire_con,
-                                         get_current_date_minus_hours)
+                                         get_diff_current_date)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


sduenas (approved):
LGTM

PR #586, opened 2019-03-22
[stackexchange] Handle missing `answer_tags`

valeriocos:
This code handles enriched items not having `answer_tags` field. Thus, it prevents the error KeyError: 'answer_tags'.

### DIFF
--- a/grimoire_elk/enriched/stackexchange.py
+++ b/grimoire_elk/enriched/stackexchange.py
@@ -234,7 +234,7 @@ class StackExchangeEnrich(Enrich):
                     rich_answer = self.get_rich_item(answer,
                                                      kind='answer',
                                                      question_tags=item['data']['tags'])
-                    if rich_answer['answer_tags']:
+                    if 'answer_tags' in rich_answer:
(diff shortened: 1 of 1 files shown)


sduenas:
Can you give me an example of this? Why that field is not available? Is StackExchange changing something?

valeriocos:
The field `answer_tags` is created during the [enrichment](https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/stackexchange.py#L195), it contains the answers that have been tagged. Later [here](https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/stackexchange.py#L237): this data is included into `answers_tags`.

The error is caused because the code tries to access to a key in the dict that doesn't exist. More code (`eitem['answer_tags'] = []` [here](https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/stackexchange.py#L195) ) can be added to the PR to ensure that `rich_answer['answer_tags']` always exists.

sduenas (approved):
LGTM

PR #587, opened 2019-03-22
[enriched] Apply retry mechanism HTTP 504 errors

valeriocos:
This code allows to extend the retry mechanism to the 504 HTTP errors.


zhquan:
LGTM

PR #588, opened 2019-03-22
[jira] Handle missing fields during the enrichment

valeriocos:
This code handle missing fields that are used to create the
enriched items. In particular, it checks that the following
fields exist:
- issue.fields.creator
- issue.fields.assignee
- issue.fields.resolution


sduenas (approved):
LGTM

PR #589, opened 2019-03-24
[jira] Handle multiple type values in `release` attribute during enrichment

valeriocos:
This code adds the mapping to the Jira enriched items to avoid ES mapping errors on the attribute `release`, which may be wrongly detected as date.


sduenas (approved):
LGTM

PR #590, opened 2019-03-25
[enriched] Fix `Too many dynamic script compilations` in git enrich

valeriocos:
This code fixes the ElasticSearch error `Too many dynamic script compilations` which occurs on the painless script used to add branch information. The script is modified to use the `params` attribute, that allows to compiles and stores the script in cache, thus avoiding compiling the script over and over.


sduenas (approved):
LGTM

PR #591, opened 2019-03-25
[enrich] Prevent timeout when using ElasticSearch DSL

valeriocos:
This code adds the param `retry_on_timeout` when initializing the ElasticSearch DSL object, thus preventing Gateway Timeout.


sduenas (approved):
LGTM

PR #593, opened 2019-03-26
[elk] Handle filter-classified param

valeriocos:
This code allows to pass the value of the filter-classifed param defined in the setup.cfg to Perceval. The methods `elk.feed_backend` and `elastic.feed` have been refactored to handle the new parameter.


sduenas (approved):
LGTM

PR #594, opened 2019-03-27
[finosmeetings] Change project value to cm_title

valeriocos:
This code sets the project value to the cm_title, which was previously cm_program.


sduenas (approved):
LGTM

PR #601, opened 2019-04-02
[enrich-git] Handle rev-list command failures

valeriocos:
Calls for the git command `rev-list` are wrapped in try..except blocks to deal with production environments which use a Git version not supporting this command.


sduenas (approved):
LGTM

PR #603, opened 2019-04-05
Update row with missing aggregatable value

alpgarcia:


dicortazar:
This looks to me! Thanks a lot :)

PR #604, opened 2019-04-05
[enriched-git] Fix handling rev-list command error

valeriocos:
The try..except block including the rev-list call in the method `add_commit_branches` is extended since the call returns a generator, thus the result of the call is calculated afterwards.

Signed-off-by: Valerio Cosentino valcos@bitergia.com


sduenas (approved):
LGTM

But please, sign the commit.

PR #605, opened 2019-04-08
[elastic_items] Handle commas in filter-raw

valeriocos:
This code allows to handle commas defined in the values of the filter-raw param within the projects.json. To achive this, the split match is changed to include also the str `data.`.

### DIFF
--- a/grimoire_elk/elastic_items.py
+++ b/grimoire_elk/elastic_items.py
@@ -25,12 +25,16 @@
 
 import json
 import logging
+import re
 
 from .enriched.utils import get_repository_filter, grimoire_con
 from .elastic_mapping import Mapping
 
 HEADER_JSON = {"Content-Type": "application/json"}
 MAX_BULK_UPDATE_SIZE = 1000
+
+FILTER_DATA_ATTR = 'data.'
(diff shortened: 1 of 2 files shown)


grimoire_elk/elastic_items.py
```diff
@@ -32,6 +32,9 @@
 HEADER_JSON = {"Content-Type": "application/json"}
 MAX_BULK_UPDATE_SIZE = 1000
 
+FILTER_DATA_ATTR = 'data.'
+FILTER_SEPARATOR = ',' + FILTER_DATA_ATTR
```
sduenas: I see a problem with is. People will forget about writing the next filter right after the comma. Some will write it after the comma and some other will leave one or several space characters.

sduenas (commented):
Is there any way to improve what I wrote regarding the `,` and `data.`?

grimoire_elk/elastic_items.py
valeriocos: It sounds reasonable, please @sduenas check the PR again. In a nutshell, a regular expression (,\s*data.) is used to split the filter. Tests have been added accordingly.

sduenas (approved):
LGTM

PR #606, opened 2019-04-08
[elk] Handle empty sh uuid attributes

valeriocos:
This code allows to ignore the uuid attributes which contain empty values, thus avoiding ES errors when inserting items with empty `_id`.


sduenas (approved):
LGTM

PR #607, opened 2019-04-09
[enriched-gerrit] Use grimoirelab-toolkit calls

valeriocos:
This code replaces datetime, time and dateutil.parser calls to equivalent calls of grimoirelab-toolkit, thus reducing possible point of failures due to duplicate code.


sduenas (approved):
LGTM

PR #608, opened 2019-04-10
[enriched-jira] Add comments to enriched index

valeriocos:
This code includes Jira comments to the enriched index. To do this, changes have been made to allow the cohesistance of both comments and issues in the same index, following an approach similar to meetup and stackexchange enrichers. Thus, enriched issues and commits have different ids, the former is the issue
ids, while the latter is a str concatenating the issue id and the comment id.

### DIFF
--- a/grimoire_elk/enriched/jira.py
+++ b/grimoire_elk/enriched/jira.py
@@ -22,12 +22,18 @@
 
 import logging
 
-from datetime import datetime
+from grimoirelab_toolkit.datetime import (datetime_utcnow,
+                                          str_to_datetime)
 
-from .enrich import Enrich, metadata
+from .enrich import Enrich, metadata, SH_UNKNOWN_VALUE
 from ..elastic_mapping import Mapping as BaseMapping
 
 from .utils import get_time_diff_days
+
+
+MAX_SIZE_BULK_ENRICHED_ITEMS = 200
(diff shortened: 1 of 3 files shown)


sduenas (approved):

grimoire_elk/enriched/jira.py
```diff
@@ -282,11 +340,102 @@ def get_rich_item(self, item):
...
+        for comment in comments:
+            ecomment = {}
+
+            for f in self.RAW_FIELDS_COPY:
+                ecomment[f] = eitem[f]
+
+            # Copy data from the enriched issue
+            ecomment['project_id'] = eitem['project_id']
+            ecomment['project_key'] = eitem['project_id']
+            ecomment['project_name'] = eitem['project_id']
```
sduenas: All these values should have the same value?

sduenas (commented):
I'm not sure if `project_id`, `project_key` and `project_name` should have the same value. In this case the value of `project_id` field. Is this the case?

I think the issue has different values for these fields.

valeriocos:
my fault, copy/paste and i didn't replace the values. thank you for pointing this out, i'll modify it asap

valeriocos:
@sduenas please re-check, thanks!

sduenas (approved):
LGTM

PR #609, opened 2019-04-10
[elk] Remove `get_items_from_uuid` method

valeriocos:
This code removes the method `get_items_from_uuid` which doesn't seem to be used neither in ELK nor in other components (e.g., sirmordred).


zhquan (approved):
LGTM

PR #610, opened 2019-04-11
[ELK] Simplify logic to assing project to enriched items

valeriocos:
This PR proposes to simplify the logic to assign project to enriched items. Currently the assignment is based on the logic of the method `get_project_repository` redefined in each enricher, plus a fallback option based on the inspection of the projects.json, which is passed from mordred to ELK. The new approach passes to ELK also the repo (`projects_json_repo`) in the projects.json (which includes the perceval params plus optionally other params such as filter-raw, filtera-raw-prefix, blacklist-ids, proper of some specific enrichers) currently analyzed. Thus, the assignment of the project is done by looking for a perfect match between the value of the `projects_json_repo` in the projects.json, in case the `projects_json_repo` is None, the previous logic is used.

This PR should be reviewed together with: https://github.com/chaoss/grimoirelab-sirmordred/pull/292

### DIFF
--- a/grimoire_elk/elastic_items.py
+++ b/grimoire_elk/elastic_items.py
@@ -39,7 +39,7 @@ FILTER_SEPARATOR = r",\s*%s" % FILTER_DATA_ATTR
 logger = logging.getLogger(__name__)
 
 
-class ElasticItems():
+class ElasticItems:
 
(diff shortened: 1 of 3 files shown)


sduenas (approved):
LGTM

PR #611, opened 2019-04-11
[raw-jira] Update raw mapping to avoid immense term errors

valeriocos:
This PR modifies the raw mappings to handle comment bodies longer than the maximum length allowed for keywords. Thus, comment bodies are mapped as text.


sduenas (approved):
LGTM

PR #612, opened 2019-04-11
[elastic] Check presence of `properties` in mappings

valeriocos:
This code checks the presence of the attribute `properties` in an enriched index. This check is needed to avoid errors when extracting properties from empty indexes.


sduenas (approved):
LGTM

PR #613, opened 2019-04-12
[enriched-utils] Fix last enrich date calculation

valeriocos:
This PR fixes the last enrich date calculation. The previous logic was calculating the minimum value between the last enrich global date (the one obtained before starting the enrichment process for all repos of a given data source) and the last enrich date for items of a given origin. If the latter date was empty (i.e., no items of that origin in the index), the last enrich date was set to the last enrich global date, thus only the items of the origin with a timestamp greater than the last enrich global date where processed. Now the last enrich date is set None if no items for a given origin are present in the enriched index.


sduenas (approved):
LGTM

PR #614, opened 2019-04-14
[enriched-bugzillarest] Change logic to calculate `timeopen_days`

valeriocos:
This code changes the logic to assess the number of days a bug is open. It relies on the attribute `is_open`, if the latter exists and is False, the `timeopen_days` is the diff between the creation time and the date of the last change, in all other cases the `timeopen_days` is set to the diff between the creation time and the date when the enriched item was processed by ELK.


sduenas (approved):
LGTM

PR #618, opened 2019-04-27
[enriched-functest] Handle no float duration times

valeriocos:
This PR changes the way of assessing the duration time of tests. It relies on `start_date` and `stop_date`, which are used to calculate the duration of a test. The result is stored in the enriched attribute `duration`. To ensure compatibility, the old way (based on the `duration` attribute included in `details`) is stored in the enriched attribute `duration_from_api`.

### DIFF
--- a/grimoire_elk/enriched/functest.py
+++ b/grimoire_elk/enriched/functest.py
@@ -21,6 +21,7 @@
 #
 
 import logging
+import re
 
 from .enrich import Enrich, metadata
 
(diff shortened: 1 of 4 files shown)


sduenas (changes requested):
The PR looks ok but please update the CSV file with the fields included.

valeriocos:
@sduenas CSV file updated, please have a look, thanks.

sduenas (approved):
LGTM

PR #619, opened 2019-04-27
[enriched-study-ceres-aoc] Limit search size to 500

valeriocos:
This PR limits the number of items returned by paginated searches to 500. This change should reduce the risk of getting `BytesStreamOutput cannot hold more than 2GB of data` errors when reading data from ElasticSearch.


sduenas (approved):
LGTM

PR #620, opened 2019-04-28
[enriched-discourse] Force type of score attribute to float

valeriocos:
This code forces the type of the score attribute to float, thus preventing ElasticSearch to thrown errors due to the assignment of different types to scores (`illegal_argument_exception`).


sduenas (approved):
LGTM

PR #623, opened 2019-05-09
[enriched-gerrit] Handle gerrit items from multiple servers

valeriocos:
This code changes the way the ids of the enriched items are composed to allow the existence of items from different gerrit servers in the same enriched index.


sduenas (approved):
LGTM

PR #627, opened 2019-05-13
[enriched-jira] Handle enriched items from multiple jiras

valeriocos:
This code modifies the way of creating ids (which now include the `uuid` of the corresponding raw items) for enriched items to handle data from multiple jira instances.


sduenas (approved):
LGTM

PR #628, opened 2019-05-20
[raw-slack] Remove channel_info.previous_names attribute

valeriocos:
This code prevents to index the field `data.channel_info.previous_names`,
thus avoiding mapping_parsing_exception.


sduenas (approved):
LGTM

PR #630, opened 2019-05-22
[enriched-confluence] Add email data to SH identity

sduenas:
In some cases, Confluence provides users email addresses. This patch adds that value to the SortingHat indentity.


valeriocos (approved):
LGTM, thanks @sduenas

PR #632, opened 2019-05-23
[schema] Specify where each author field come from

alpgarcia:
Person related fields can be retrieved from the original
data source or rom SortingHat profile. This commit adds
where each of them comes from.

Signed-off-by: Alberto Pérez García-Plaza <alpgarcia@bitergia.com>


sduenas (approved):
LGTM

PR #634, opened 2019-05-25
[enriched-github] Filter review comments from ghost users

valeriocos:
This code filters the review comments left by users that are not present anymore on github. Test data and test have been changed accordingly.


sduenas (approved):
LGTM

PR #636, opened 2019-05-29
[dependencies] Pin urllib3 and requests dependencies

valeriocos:
This code fixes the urllib3 version used by ELK to 1.22, since in the later versions urllib3.exceptions.SSLError exceptions may be thrown due to CERTIFICATE_VERIFY_FAILED. Furthermore, it also sets the version of requests to 2.18.2, which is compatible with urllib3 1.22.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,8 @@
 PyMySQL
 elasticsearch==6.3.1
 elasticsearch-dsl==6.3.1
-requests>=2.7.0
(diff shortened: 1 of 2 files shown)


sduenas (changes requested):
The commit message is wrong. It talks about Perceval. As I commented in Perceval, we should pin requests version too.

sduenas (approved):
LGTM

PR #637, opened 2019-05-30
Fix elasticsearch.exceptions.SSLError: ConnectionError

valeriocos:
This PR updates the statements that initialize the Elasticsearch objects (imported from the library elasticsearch) to use `RequestsHttpConnection` as connection class, thus avoiding `elasticsearch.exceptions.SSLError: ConnectionError` exceptions. Furthermore the dependencies of urllib3 and requests are also updated, since urllib3 1.22 is affected by a high severity security vulnerability.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ before_install:
   - pip install redis
   - pip install PyMySQL
   - pip install httpretty==0.8.6
-  - pip install elasticsearch==6.1.1
-  - pip install elasticsearch-dsl==6.1.0
+  - pip install elasticsearch==6.3.1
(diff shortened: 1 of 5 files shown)


sanacl:
@valeriocos why you are not using the urllib3 version `1.25.2`?

valeriocos:
Because it's not in the range declared at: https://github.com/kennethreitz/requests/blob/v2.21.0/setup.py#L47

I have tried to use `requests` 2.22.0 but travis was complaining when running the tests, so I used the latest version available

sduenas (approved):
LGTM

PR #638, opened 2019-06-04
[raw-slack] Fix illegal_argument_exception for data.blocks.elements.text

valeriocos:
This code handles `illegal_argument_exception` errors on the field `data.blocks.elements.text` which prevent the insertion of slack items. Tests data has beeen changed accordingly.


sduenas (approved):
LGTM

PR #639, opened 2019-06-04
[raw-bugzillarest] Handle description longer than 32766 chars

valeriocos:
This code sets the mapping for bug descriptions to text, thus allowing to prevent immense term exceptions. Tests data has been modified accordingly.


alpgarcia:
This should work and LGTM.

Not sure if we all are aware of the difference between indexed and stored fields [1]. In ES, afaik, all fields are stored within the `_source` field (enabled by default). If we have other smaller text fields in the RAW index we want to retrieve without accessing the `_source` field, we could configure them to be `stored` so they could be retrieved independently.

In the general case I guess we are only querying the index or retrieving entire documents for enrichment, so we probably don't need it. Anyway I'd say it is an interesting reading when defining ES mapping so I just want to make sure we are all aware of these config options to avoid being the single point of failure if I misunderstand something :P


[1] https://www.elastic.co/guide/en/elasticsearch/reference/6.1/mapping-store.html

alpgarcia (approved):
LGTM, thanks!

PR #640, opened 2019-06-04
[enriched-enrich] Reduce number of log messages on identity retrieval

valeriocos:
This code reduces the number of messages logged when retrieving identity information from SortingHat. To this purpose, a check is added to verify that the identity returned by SortingHat is not None, and log messages missing identities and invalid values are set to `debug` level.


sduenas (approved):
LGTM

PR #641, opened 2019-06-10
[raw-jira] Handle data.fields.environment in raw indexes

valeriocos:
This code indexes the data.fields.environment attribute as text value, thus avoiding immense term exceptions. Test data has been updated accordingly.


sduenas (approved):
LGTM

PR #643, opened 2019-06-25
[elk] Use grimoirelab-toolkit datetime functions

valeriocos:
This PR removes the function `unixtime_to_datetime` in enriched/utils.py, which was a copy of `datetime.unixtime_to_datetime` in grimoirelab-toolkit. Furthermore, it replaces the use of datetime and dateutil calls with their equivalent in grimoirelab-toolkit. Other enrichers (e.g., bugzilla) are still using direct datetime and dateutil calls, however they can be addressed in a new PR.

### DIFF
--- a/grimoire_elk/elastic.py
+++ b/grimoire_elk/elastic.py
@@ -29,9 +29,10 @@ from time import time
 
 import requests
 
+from grimoirelab_toolkit.datetime import unixtime_to_datetime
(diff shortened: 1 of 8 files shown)


alpgarcia (approved):
LGTM

sduenas (approved):
LGTM

PR #644, opened 2019-06-25
[enriched-meetup] Store created, time, updated as date

valeriocos:
This code converts the attributes `meetup_time`, `meetup_created` and `meetup_updated` to a date
and store it in isoformat. Tests have been added accordingly.

### DIFF
--- a/grimoire_elk/enriched/meetup.py
+++ b/grimoire_elk/enriched/meetup.py
@@ -25,7 +25,7 @@ import logging
 
 from .enrich import Enrich, metadata
 
-from .utils import unixtime_to_datetime
+from grimoirelab_toolkit.datetime import unixtime_to_datetime
 from ..elastic_mapping import Mapping as BaseMapping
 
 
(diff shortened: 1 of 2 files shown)


grimoire_elk/enriched/meetup.py
```diff
@@ -183,6 +183,9 @@ def get_rich_item(self, item):
         eitem['num_rsvps'] = len(rsvps)
         eitem['num_comments'] = len(event['comments'])
 
+        if eitem['meetup_created']:
+            eitem['meetup_created'] = unixtime_to_datetime(eitem['meetup_created'] / 1000).isoformat()
+
```
alpgarcia: I know this comment could be out of the scope of the issue I notified, but I realized some of the `copy_fields` here are in fact dates too. Specifically: `created`, `updated` and `time`. Would it make sense to have a separate list for them and convert them to dates (instead of copying them and then updating them)?

alpgarcia (commented):
@valeriocos please have a look at my comment and let me know if it makes sense to you.

Thanks!

grimoire_elk/enriched/meetup.py
valeriocos: Yes, sure! I can update the PR

valeriocos: PR updated

valeriocos: PR updated, now two different loops are used to copy the target fields, one of them focuses on the fields to be converted to datetime.

alpgarcia (approved):
LGTM, thanks!

PR #645, opened 2019-06-25
[enriched-enrich] Decrease max keyword size

valeriocos:
This PR limits the number of chars for a string mapped as keyword to 1000 chars. The change is needed to to avoid max_bytes_length_exceeded_exception errors, which may occur when the UTF8 encoding is longer than the max length allowed for keyword attributes (32766 bytes). Furthermore, this PR updates the name of the variable `KEYWORD_MAX_SIZE` to `KEYWORD_MAX_LENGTH`, which seems to be more appropiated.


sduenas (approved):
LGTM

PR #646, opened 2019-06-26
Added check so that program does not break if aliases is None

alangsto:
Signed-off-by: Alison Langston <alangsto@wellesley.edu>

I encountered an issue while running grimoirelabs/installed with elasticsearch and kibana hosted on aws instances. I did not encounter this same issue while running grimoirelabs/full locally. 

After changing the line of code in the pull request, the program was able to run again. 

The code changed handles if aliases is None, so that the program can continue successfully.


valeriocos:
Thank you @alangsto for the PR.

I'll try to replicate the error.

Could you share your setup.cfg? Maybe the problem is related to some indexes names, which are clashing with the [default aliases names](https://github.com/chaoss/grimoirelab-sirmordred/blob/master/aliases.json)? ElasticSearch doesn't allow to create an alias with a name already used by an index.

PS Please remove the file `.DS_Store` :)

alangsto:
Hi @vaeriocos 

I actually haven't changed anything in the files that are created in the grimoirelab/installed version except for the urls/passwords/usernames in the infra.cfg. 

I found the error I was getting very strange (the error was that I couldn't iterate over NoneType, because aliases was None), because I had been using various images from grimoirelabs for the past month and had not run into this problem until I had moved my elasticsearch, kibana, and sql database onto aws instances. In any case, adding this line of code fixed my problem, and now the dashboards are working. I'll share my config files with you, although I'm not sure how easy this error would be to reproduce unless you are running it with the exact same type of elasticsearch/kibana/sql database. 

This is the only config file I changed.
infra.cfg:
```
[general]
# Directory for logs
logs_dir = /logs
# File with aliases definitions
aliases_file = /aliases.json
# File with menu defintions
menu_file = /menu.yaml

[projects]
projects_file = /projects.json

[es_collection]
# Elasticsearch for data collection (raw indexes)
url = https://xxx:xxx@847c0596a6326412c0313eb32d88595a.us-east-1.aws.found.io:9243
user =
password =

[es_enrichment]
# Elasticsearch for enriched data
url = https://xxx:xxx@847c0596a6326412c0313eb32d88595a.us-east-1.aws.found.io:9243
user =
password =

[sortinghat]
# Infrastructure for SortingHat (MariaDB/MySQL database)
host = ui0tj7jn8pyv9lp6.cbetxkdyhwsb.us-east-1.rds.amazonaws.com
user = xxx
password = xxx
database = xxx

# Use organizations file
load_orgs = true
orgs_file = /orgs.json

# Identities file in GrimoireLab format
#identities_file = [/identities.yaml]
#identities_format = grimoirelab

[panels]
# Kibitter / kibana entry point
kibiter_url = https://xxx:xxx@66f663f015aecb785655faf0aae52d04.us-east-1.aws.found.io:9243
```

valeriocos:
Thank you for the explanation @alangsto . I'll have a look at it (maybe with the help of @jgbarah) and keep you informed.

Please squash the commits in one :)

valeriocos:
Hi @alangsto I'm not able to reproduce the issue you report. The current code ensures that all indexes (raw and enriched ones) have an alias. The alias name is either read from a file (which path is defined in the setup.cfg) or set automatically by the code (see pointers to the code below). Furthermore, the code that sets the alias is triggered only if aliases exist (https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/elastic.py#L104)

**pointers:**
- https://github.com/chaoss/grimoirelab-sirmordred/blob/master/sirmordred/task_collection.py#L65
- https://github.com/chaoss/grimoirelab-sirmordred/blob/master/sirmordred/task_enrich.py#L74
- https://github.com/chaoss/grimoirelab-sirmordred/blob/master/sirmordred/config.py#L160

The change in the condition you propose can just improve the code, so the PR could be accepted. WDYT @sduenas ?

alangsto:
Hey @valeriocos, 
Thank you for the extra info! 

The problem that I was having was that the aliases here [https://github.com/chaoss/grimoirelab-elk/blob/7efcca1f8293c5b7207134c9f0c3937c4ebab123/grimoire_elk/elastic.py#L70](url) were different than the aliases being referenced here [https://github.com/chaoss/grimoirelab-elk/blob/7efcca1f8293c5b7207134c9f0c3937c4ebab123/grimoire_elk/elastic.py#L140](url).

Again, I'm not sure why this was happening and it is not happening anymore, although it did occur and the aliases in line 140 were None.

valeriocos:
Thank you @alangsto , so in this case we can close the PR. WDYT @sduenas?

sduenas:
@valeriocos I think we can merge it too. It makes code more robust, IMHO.

PR #647, opened 2019-06-27
[enriched] Add text mapping

valeriocos:
This PR add text mappings for some enriched attributes: 
- bugzilla: `summary_analyzed` and `main_description_analyzed`
- bugzillarest:  `summary_analyzed` and `main_description_analyzed`
- discourse: `question_title_analyzed`
- gerrit: `approval_description_analyzed` and `comment_message_analyzed`
- jira: `main_description_analyzed`
- mediawiki: `revision_comment_analyzed`
- mozillaclub: `Event_Description_Analyzed`, `Event_Creations_Analyzed`, `Feedback_from_Attendees_Analyzed` and `Your_Feedback_Analyzed`
- redmine: `description_analyzed` and `subject_analyzed`

Furthermore, it also promotes the use of the grimoirelab-toolkit datetime functions (instead of calls to the datetime/dateutil libraries)

For consistency reasons, the param `index` for all text mappings is now set explictly to `true`

### DIFF
--- a/grimoire_elk/enriched/askbot.py
+++ b/grimoire_elk/enriched/askbot.py
@@ -49,10 +49,12 @@ class Mapping(BaseMapping):
         {
             "properties": {
                 "author_badges": {
-                  "type": "text"
+                  "type": "text",
+                  "index": true
                 },
                 "summary": {
-                  "type": "text"
+                  "type": "text",
+                  "index": true
                 },
                 "id": {
                   "type": "keyword"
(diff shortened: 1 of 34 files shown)


alpgarcia (approved):
LGTM. Just a minor comment: there are cases with `"index": true` and others without. It defaults to true, so they both have the same effect. I would go for using always the same mapping configuration (either explicitly or implicitly declaring `index` value) for clarity.

valeriocos:
Yes, but as you know ELK lacks of consistent choices for historical reasons and in the past several attempts to improve the overall code quality have been blocked.

I have just added a commit that explicitly sets the index value to `true` for text fields.

@alpgarcia feel free to review it again

alpgarcia:
Thanks @valeriocos ! LGTM, merging...

PR #648, opened 2019-06-28
[elk] Return error msg from `feed_backend`

valeriocos:
This PR modifies the `feed_backend` method to return an error msg. This information is then consumed by Mordred.

### DIFF
--- a/grimoire_elk/elk.py
+++ b/grimoire_elk/elk.py
@@ -32,6 +32,7 @@ from dateutil import parser
 
 from arthur.common import Q_STORAGE_ITEMS
 from perceval.backend import find_signature_parameters, Archive
+from perceval.errors import RateLimitError
 from grimoirelab_toolkit.datetime import datetime_utcnow
 
 from .elastic_mapping import Mapping as BaseMapping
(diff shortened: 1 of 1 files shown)


grimoire_elk/elk.py
```diff
@@ -214,13 +216,19 @@ def feed_backend(url, clean, fetch_archive, backend_name, backend_params,
 
             ocean_backend.feed(**params)
 
+    except RateLimitError as ex:
+        logger.error("Error feeding ocean from %s (%s): %s", backend_name, backend.origin, ex)
```
sduenas: Shouldn't this message be different? Like `Feeding cancelled from {} ({}) due to rate limit exceeded; wait {} seconds to restart the process`

sduenas (commented):
Check if my comment makes sense.

grimoire_elk/elk.py
valeriocos: I didn't change it to avoid modifying existing things. If I'm not wrong ops team is used to look for the msg `Error feeding ocean from` in the logs. Maybe we can involve @zhquan and @sanacl in the loop.

sanacl: You can change it @valeriocos . All we need to know is the new message to be tracked.

valeriocos: OK, let's try a solution in the middle, we change but keep the similar shape:
`"Error feeding ocean from %s (%s): rate limit exceeded"`

WDYT?

sduenas (approved):
LGTM

PR #652, opened 2019-07-02
[raw-bugzillarest] Set text mapping type for changes attributes

valeriocos:
This code sets the mapping for the attributes `data.history.changes.removed` and `data.history.changes.added` to text.

Tests data have been changed accordingly.


alpgarcia (approved):
I checked the old mapping in action and it was defined as `keyword` (see below). IMHO, there is no reason to have these long text fields stored as keyword, so adding the mapping as @valeriocos should solve the problem and also makes more sense. 

```
              "history": {
                "properties": {
                  "changes": {
                    "properties": {
                      "added": {
                        "type": "keyword"
                      },
                      "attachment_id": {
                        "type": "long"
                      },
                      "comment": {
                        "properties": {
                          "already_wrapped": {
                            "type": "keyword"
                          },
                          "bug_id": {
                            "type": "keyword"
                          },
                          "bug_when": {
                            "type": "keyword"
                          },
                          "comment_id": {
                            "type": "keyword"
                          },
                          "edit_count": {
                            "type": "keyword"
                          },
                          "extra_data": {
                            "type": "keyword"
                          },
                          "is_markdown": {
                            "type": "keyword"
                          },
                          "isprivate": {
                            "type": "keyword"
                          },
                          "thetext": {
                            "type": "keyword"
                          },
                          "type": {
                            "type": "keyword"
                          },
                          "who": {
                            "type": "keyword"
                          },
                          "work_time": {
                            "type": "keyword"
                          }
                        }
                      },
                      "comment_count": {
                        "type": "long"
                      },
                      "comment_id": {
                        "type": "long"
                      },
                      "field_name": {
                        "type": "keyword"
                      },
                      "removed": {
                        "type": "keyword"
                      }
                    }
                  },
```

PR #654, opened 2019-07-08
[mediawiki] Fix grimoire_creation_date calculation

zhquan:
This code changes the way of calculating the `grimoire_creation_date`, which is now derived
from the timestamp of each revision. Revisions can be of 2 types: `iscreation` (if parent is None) or `isrevision` (if parent isn't None).

Tests have been provided accordingly.

### DIFF
--- a/grimoire_elk/enriched/mediawiki.py
+++ b/grimoire_elk/enriched/mediawiki.py
@@ -180,20 +180,21 @@ class MediaWikiEnrich(Enrich):
 
             erevision["url"] = erevision["url"].replace(" ", "_")
             erevision["iscreated"] = 0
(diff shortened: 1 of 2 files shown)


valeriocos (approved):
LGTM, thanks @zhquan

alpgarcia (approved):
Before this change `grimoire_creation_date` was copied from `metadata__updated_on`. Besides, `creation_date` was only set for pages, not for revisions.

This PR addresses both problems by setting `creation_date` for both, pages and revisions (in fact pages and revisions are the same, we just consider the very first revision as page creation), and using it to get `grimoire_creation_date`.

LGTM, thanks!

PR #658, opened 2019-07-15
[enriched-github] Set roles to allow refresh identities

valeriocos:
This code sets the variables roles to merged_by_data, user_data, assignee_data, thus allowing the refresh
of sortinghat information for the corresponding identities.


mafesan:
Thank you @valeriocos ! LGTM, I have tested this change and it works as expected, refresing the fields when an identity changes in SortingHat.

mafesan (approved):

PR #661, opened 2019-07-16
[enriched-jira] Track creator, assignee, reporter in issues

valeriocos:
For every issue, this code creates 3 enriched items which contain creator, assignee and reporter data. The enriched items have type issue and one of the following user_type: creator, assignee and reporter. This change eases the way of counting contributions of each user.

Tests have been added accordingly, schema has been updated.

### DIFF
--- a/grimoire_elk/enriched/jira.py
+++ b/grimoire_elk/enriched/jira.py
@@ -35,7 +35,6 @@ MAX_SIZE_BULK_ENRICHED_ITEMS = 200
 ISSUE_TYPE = 'issue'
 COMMENT_TYPE = 'comment'
 
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


alpgarcia (changes requested):
After talking to @valeriocos we decided to include `creator_*`, `reporter_*` and `assignee_*` fields for all user types, keeping the index retro-compatible with Sigils panels. It also eases the way of computing some metrics like, for instance, showing number of issues created by a given person and the number of people that were assigned to those issues.

valeriocos:
@alpgarcia I have just updated the PR, please have a look

alpgarcia (approved):
I've tested it locally and it works fine.

These changes need to update Sigils Jira panels to filter by `author_type: creator` in order to keep them working as they were created (to track data from the point of view of issue creators)

PR #663, opened 2019-07-19
[enriched-enrich] Handle errors in demography study

valeriocos:
This code allows to handle errors that may pop up during the execution of the demography study. The
update_by_query underlying the study is executed with the params `wait_for_completion=true` and
`conflicts=proceed`. The former waits that the updates have finished, while the latter prevents
to stop the study in case of a version conflicts in the documents updated. Furthermore, the request
issuing the update_by_query is wrapped in a try/except block to ignore Retry errors, which may occur when ElasticSearch is not available.


zhquan:
LGTM

sduenas (approved):
LGTM

PR #668, opened 2019-07-22
[tests] Increase waiting time for demography study

valeriocos:
This code increase the waiting time to check that the demography study correctly set the corresponding
fields.


sduenas (approved):
LGTM

PR #664, opened 2019-07-19
[graal] Add support of Graal's CoCom Backend to ELK [File-level + Study] (WIP)

inishchith:
Address: https://github.com/inishchith/gsoc-graal/issues/8

**TODO:**
- [x] Add tests
- [x] Update docstrings
- [x] Update commit description

**NOTE:**
- The failing test is due to `Graal` module not being a part of `requirements.txt`(dependencies) and is a part of this PR just to check the coverage of new additions.

Signed-off-by: inishchith <inishchith@gmail.com>

PR #669, opened 2019-07-23
[graal] Add support of Graal's CoLic Backend to ELK [File-level + Study] (WIP) 

inishchith:
Address: https://github.com/inishchith/gsoc-graal/issues/9

**TODO:**
- [x] Add tests
- [x] Update docstrings
- [x] Update commit description

**NOTE:**
- The failing test is due to `Graal` module not being a part of `requirements.txt`(dependencies) and is a part of this PR just to check the coverage of new additions.

Signed-off-by: inishchith <inishchith@gmail.com>

### DIFF
--- a/grimoire_elk/enriched/colic.py
+++ b/grimoire_elk/enriched/colic.py
@@ -1,0 +1,346 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2015-2019 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+#
+# Authors:
+#   Nishchith Shetty <inishchith@gmail.com>
+#
+
+import logging
+from dateutil.relativedelta import relativedelta
+
+from elasticsearch import Elasticsearch as ES, RequestsHttpConnection
+from .enrich import (Enrich,
+                     metadata)
+from .graal_study_evolution import (get_to_date,
+                                    get_unique_repository,
+                                    get_files_at_time)
+
+from grimoirelab_toolkit.datetime import (str_to_datetime,
+                                          datetime_utcnow)
+from grimoire_elk.elastic import ElasticSearch
+
+MAX_SIZE_BULK_ENRICHED_ITEMS = 200
+
+logger = logging.getLogger(__name__)
+
+
+class ColicEnrich(Enrich):
+
+    def __init__(self, db_sortinghat=None, db_projects_map=None, json_projects_map=None,
+                 db_user='', db_password='', db_host=''):
+        super().__init__(db_sortinghat, db_projects_map, json_projects_map,
+                         db_user, db_password, db_host)
+
+        self.studies = []
+        self.studies.append(self.enrich_colic_analysis)
+
+    def get_identities(self, item):
+        """ Return the identities from an item """
+        identities = []
+
+        return identities
+
+    def has_identities(self):
+        """ Return whether the enriched items contains identities """
+
+        return False
+
+    def get_field_unique_id(self):
+        return "id"
+
+    def get_licensed_files(repository_url, to_date):
+        """ Retrieve all the licensed files until the to_date, corresponding
+        to the given repository.
+        """
+
+        query_licensed_files = """
+        {
+            "size": 0,
+            "aggs": {
+                "1": {
+                    "cardinality": {
+                        "field": "file_path"
+                    }
+                }
+            },
+            "query": {
+                "bool": {
+                    "must": [{
+                        "match_phrase": {
+                            "has_license": {
+                                "query": 1
+                            }
+                        }
+                    },
+                    {
+                        "match_phrase": {
+                            "origin": {
+                                "query": "%s"
+                            }
+                        }
+                    },
+                    {
+                        "range": {
+                            "metadata__updated_on": {
+                                "lte": "%s"
+                            }
+                        }
+                    }]
+                }
+            }
+        }
+        """ % (repository_url, to_date)
+
+        return query_licensed_files
+
+    def get_copyrighted_files(repository_url, to_date):
+        """ Retrieve all the copyrighted files until the to_date, corresponding
+        to the given repository.
+        """
+
+        query_copyrighted_files = """
+        {
+            "size": 0,
+            "aggs": {
+                "1": {
+                    "cardinality": {
+                        "field": "file_path"
+                    }
+                }
+            },
+            "query": {
+                "bool": {
+                    "must": [{
+                        "match_phrase": {
+                            "has_copyright": {
+                                "query": 1
+                            }
+                        }
+                    },
+                    {
+                        "match_phrase": {
+                            "origin": {
+                                "query": "%s"
+                            }
+                        }
+                    },
+                    {
+                        "range": {
+                            "metadata__updated_on": {
+                                "lte": "%s"
+                            }
+                        }
+                    }]
+                }
+            }
+        }
+        """ % (repository_url, to_date)
+
+        return query_copyrighted_files
+
+    def extract_modules(self, file_path):
+        """ Extracts module path from the given file path """
+        path_chunks = file_path.split('/')
+
+        modules = []
+        for idx in range(len(path_chunks)):
+            sub_path = '/'.join(path_chunks[:idx])
+
+            if sub_path:
+                modules.append(sub_path)
+
+        return modules
+
+    @metadata
+    def get_rich_item(self, file_analysis):
+        # TODO: requires adjustments regarding category of backend used
+
+        eitem = {}
+        eitem["file_path"] = file_analysis["file_path"]
+        eitem["modules"] = self.extract_modules(eitem["file_path"])
+        eitem["copyrights"] = []
+        eitem["licenses"] = []
+        eitem["license_name"] = []
+        eitem["has_license"] = 0
+        eitem["has_copyright"] = 0
+
+        if file_analysis.get("licenses", False):
+            eitem["has_license"] = 1
+            for _license in file_analysis["licenses"]:
+                eitem["licenses"].extend(_license["matched_rule"]["licenses"])
+                eitem["license_name"].append(_license["name"])
+
+        if file_analysis.get("copyrights", False):
+            eitem["has_copyright"] = 1
+            for _copyright in file_analysis["copyrights"]:
+                eitem["copyrights"].append(_copyright["value"])
+
+        return eitem
+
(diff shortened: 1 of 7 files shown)


valeriocos (commented):
Thank you @inishchith for the PR. Overall it looks OK. I see that some code of the study is similar to the one proposed here: https://github.com/chaoss/grimoirelab-elk/pull/664. It would be possible to abstract the common logic to a class (e.g., study_evolution) and call it from cocom and colic enrichers?

A similar comment targets the file `graal_utils`, it would be possible to wrap the variables in `graal_utils` to methods, which use some variables to set up the query attributes? If this is possible, then we could include these methods in the `study_evolution` file.

WDYT?

inishchith:
@valeriocos Yes. Agree with your thoughts. Will update the PR once I've worked on it.

Thanks!

grimoire_elk/enriched/colic.py
```diff
@@ -0,0 +1,266 @@
...
+        return eitem
+
+    def get_rich_items(self, item):
+        # The real data
+        entry = item['data']
+
+        enriched_items = []
+
+        for file_analysis in entry["analysis"]:
+            eitem = self.get_rich_item(file_analysis)
```
valeriocos: Here we could have an if-else block to enrich the items depending on the category found (nomossa, scancode, etc.), WDYT? Please, have a look at this code: https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/github.py#L317 as reference

grimoire_elk/enriched/colic.py
```diff
@@ -0,0 +1,266 @@
...
+        else:
+            logger.info("%s items inserted for CoLic", str(num_items))
+
+        return num_items
+
+    def enrich_colic_analysis(self, ocean_backend, enrich_backend, no_incremental=False,
+                              out_index="colic_enrich_graal_repo", interval_months=3,
+                              date_field="grimoire_creation_date"):
+
+        logger.info("Doing enrich_repository_analysis study for index {}"
```
valeriocos: `enrich_repository_analysis` should be `enrich_colic_analysis`

grimoire_elk/enriched/colic.py
```diff
@@ -0,0 +1,266 @@
...
+
+        es_in = ES([enrich_backend.elastic_url], retry_on_timeout=True, timeout=100,
+                   verify_certs=self.elastic.requests.verify, connection_class=RequestsHttpConnection)
+        in_index = enrich_backend.elastic.index
+
+        unique_repos = es_in.search(
+            index=in_index,
+            body=get_unique_repository())
+
+        repositories = [repo['key'] for repo in unique_repos['aggregations']['unique_repos']['buckets']]
```
valeriocos: Same comment of https://github.com/chaoss/grimoirelab-elk/pull/664/files#r307905453

grimoire_elk/enriched/colic.py
```diff
@@ -0,0 +1,266 @@
...
+        num_items = 0
+        ins_items = 0
+
+        for repository_url in repositories:
+            es_out = ElasticSearch(enrich_backend.elastic.url, out_index)
+            evolution_items = []
+
+            to_date = get_to_date(es_in, in_index, out_index, repository_url)
+            to_date = to_date.replace(day=1, hour=0, minute=0, second=0)
```
valeriocos: Some comment of https://github.com/chaoss/grimoirelab-elk/pull/664/files#r307906735

grimoire_elk/enriched/colic.py
```diff
@@ -0,0 +1,266 @@
...
+            today_date = datetime_utcnow().replace(day=1, hour=0, minute=0, second=0)
```
valeriocos: Same comment of https://github.com/chaoss/grimoirelab-elk/pull/664/files#r307906232

grimoire_elk/enriched/colic.py
```diff
@@ -0,0 +1,266 @@
...
+                    index=in_index,
+                    body=get_files_at_time(repository_url, to_date.isoformat()))
+
+                licensed_files = int(licensed_files_at_time["aggregations"]["1"]["value"])
+                copyrighted_files = int(copyrighted_files_at_time["aggregations"]["1"]["value"])
+                total_files = int(files_at_time["aggregations"]["1"]["value"])
+
+                repository_name = repository_url.split("/")[-1]
+                evolution_item = {
+                    "id": to_date.strftime('%Y-%m-%dT%H:%M:%S') + "_" + repository_name + "_" + str(interval_months),
```
valeriocos: Same comment of https://github.com/chaoss/grimoirelab-elk/pull/664/files#r307908191

grimoire_elk/enriched/colic.py
```diff
@@ -0,0 +1,266 @@
...
+                    "origin": repository_url,
+                    "interval_months": interval_months,
+                    "study_creation_date": to_date.isoformat(),
+                    "licensed_files": licensed_files,
+                    "copyrighted_files": copyrighted_files,
+                    "total_files": total_files
+                }
+
+                pprint(evolution_item)
```
valeriocos: Same comment of https://github.com/chaoss/grimoirelab-elk/pull/664/files#r307909284

grimoire_elk/enriched/colic.py
```diff
@@ -0,0 +1,266 @@
...
+
+                to_date = to_date + relativedelta(months=+interval_months)
+
+            if len(evolution_items) > 0:
+                num_items += len(evolution_items)
+                ins_items += es_out.bulk_upload(evolution_items, self.get_field_unique_id())
+
+            if num_items != ins_items:
+                missing = num_items - ins_items
+                logger.error("%s/%s missing items for Study Enricher", str(missing), str(num_items))
```
valeriocos: Same comment of https://github.com/chaoss/grimoirelab-elk/pull/664/files#r307909566

grimoire_elk/enriched/colic.py
```diff
@@ -0,0 +1,266 @@
...
+            else:
+                logger.info("%s items inserted for Study Enricher", str(num_items))
+
+    def __fix_field_date(self, date_value):
```
valeriocos: I see that this method is duplicated here: https://github.com/chaoss/grimoirelab-elk/pull/664/files#diff-24d429520eb5fff2de2d44406c4d2eebR258

Would make sense to transform this method to a function and add it to https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/utils.py, WDYT?

grimoire_elk/enriched/graal_study_evolution.py
```diff
@@ -0,0 +1,280 @@
...
+                }]
+            }
+        }
+    }
+    """ % (repository_url, to_date)
+
+    return query_licensed_files
+
+
+def get_copyrighted_files(repository_url, to_date):
```
valeriocos: This function could be moved to the colic file, since it is tailored to that file, WDYT?

grimoire_elk/enriched/graal_study_evolution.py
```diff
@@ -0,0 +1,280 @@
...
+                }]
+            }
+        }
+    }
+    """ % (repository_url, to_date)
+
+    return query_files_at_time
+
+
+def get_licensed_files(repository_url, to_date):
```
valeriocos: This function could be moved to the colic file, since it is tailored to that file, WDYT?

grimoire_elk/raw/colic.py
```diff
@@ -0,0 +1,70 @@
...
+#
+# Authors:
+#   Nishchith Shetty <inishchith@gmail.com>
+#
+
+from .elastic import ElasticOcean
+from ..elastic_mapping import Mapping as BaseMapping
+
+
+class Mapping(BaseMapping):
```
valeriocos: This file is equal to https://github.com/chaoss/grimoirelab-elk/pull/664/files#diff-5e7a94ba436c87e10aff3c942a6fc3b5R28. Is it probably better to define just one GraalOcean, and update the info in the utils.py accordingly:
- https://github.com/chaoss/grimoirelab-elk/pull/669/files#diff-43cc8acd5841812504d94a0579836291R206
https://github.com/chaoss/grimoirelab-elk/pull/664/files#diff-43cc8acd5841812504d94a0579836291R207

WDYT?

valeriocos (commented):
Thank you @inishchith! Also this PR looks great. The comments are similar to the PR https://github.com/chaoss/grimoirelab-elk/pull/664, plus there are some suggestions to reduce redundant code. Let me know what you think.

grimoire_elk/enriched/colic.py
inishchith: @valeriocos I like the idea and had thought about it earlier. The primary concern is about the level of information that we can provide while they're(others) included as an option. 
I'll think over it again, look for possibilities and update this thread with more information.

Thanks for pointing it out!

grimoire_elk/enriched/colic.py
inishchith: That would be much better I feel. Thanks, I'll update the code if you agree.

grimoire_elk/enriched/graal_study_evolution.py
inishchith: My thought of both methods(`get_copyrighted_files`, `get_licensed_files`) was to have all the queries related to Graal at a single place. But I feel your point is valid and would be better if we go ahead with it!

grimoire_elk/enriched/graal_study_evolution.py
inishchith: _[ same comment as above :) ]_

grimoire_elk/raw/colic.py
inishchith: Yes. Agree with you. Thanks for pointing it out!

grimoire_elk/enriched/colic.py
valeriocos: Yes, perfect!

grimoire_elk/raw/colic.py
valeriocos: Wouldn't be better to work on a unique PR which will contain the CoCom, CoLic changes plus the modification to the utils.py file, WDYT? It will be easier to add modifications and test them

grimoire_elk/enriched/colic.py
```diff
@@ -0,0 +1,352 @@
...
+
+        eitem = {}
+
+        # entry["holders"] = file_analysis["holders"][0]["value"]
+        eitem["file_path"] = file_analysis["file_path"]
+        eitem["modules"] = self.extract_modules(eitem["file_path"])
+        eitem["copyrights"] = []
+        eitem["licenses"] = []
+        eitem["license_name"] = []
+        # eitem["owner"] = []
```
valeriocos: please delete these comments lines if not needed

valeriocos (commented):
Thanks @inishchith to address my comments. Overall the PR looks great. 
I'm not sure whether it's better to unify this PR with #664, since there are common pieces of code, and some modifications cannot be done. WDYT?

inishchith:
Thanks for the review @valeriocos.
I was thinking to open a new PR(and close these by referencing them in the newer one) by the following levels of commit:
1. Requirements 
2. CoCom Integration
3. CoLic Integration
4. Segregating common parts out
5. Tests

Which would allow us to focus on Graal integrations as a whole and testing would become much easier.
Let me know what you think :)

valeriocos:
You're welcome @inishchith and thanks for proposing the levels of commit :)

I would add a 6. point to add the csvs of the colic and cocom (as reference you can have a look at the [git.csv](https://github.com/chaoss/grimoirelab-elk/blob/master/schema/git.csv))

inishchith:
closing in reference to #672

PR #674, opened 2019-08-08
Fix outdated link to GrimoireLab tutorial

alpgarcia:
It seems we never update the link to GrimoireLab tutorial...


valeriocos (approved):
LGTM

PR #675, opened 2019-08-09
[meetup] Store `group_created` as date

valeriocos:
This code stores the attribute `group_created` as date. Furthermore, corresponding tests and CSV have been updated accordingly.

### DIFF
--- a/grimoire_elk/enriched/meetup.py
+++ b/grimoire_elk/enriched/meetup.py
@@ -235,6 +235,9 @@ class MeetupEnrich(Enrich):
                 "lon": group['lon'],
             }
 
+            if eitem['group_created']:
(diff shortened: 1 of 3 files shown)


mafesan (approved):
I have tested this PR collecting Meetup data from scratch executing both `raw` and `enrich` phases. After creating a new index pattern, I checked the type and values for `group_created` field and I confirm it appears now as type `date`. As a double-check, looking at the values for this field in actual documents, they are stored as dates now. Example:
```
"group_created": "2017-02-16T15:27:57+00:00"
```
Thanks @valeriocos! LGTM.

valeriocos:
great, thanks @mafesan !

PR #676, opened 2019-08-09
[enriched-gerrit] Add time to first review for changesets and patchsets

valeriocos:
This PR calculates the time to first review of gerrit changesets and patchsets by making a diff in days of the time when the changeset/patchset was created and the date of the first approval of type code-review not made by the author of the changeset/patchset.

Tests and CSV have been updated accordingly.


valeriocos:
An evaluation of this PR has been conducted over 100.000 reviews. 
(a) The calculation of `first time to review` only on changesets takes around 9:30 minutes.
(b) The calcuation of `first time to review` for changesets and patchsets takes around 11 minutes. Thus, the enrichment phase with (b) takes around 13.6% more time than (a) .

alpgarcia (changes requested):
It seems the code is not properly working in some cases. I tested it and found this case with negative values for most patchsets:
https://gerrit.onap.org/r/c/clamp/+/86012

Captura de pantalla de 2019-08-22 12-45-54

Although most reviews were made by a bot, we shouldn't show negative numbers. The question would be whether or not to show time to first review when the patchset was rejected (-2) by a bot.

valeriocos:
I have been checking the issue, it seems that the logic works fine, but there is something wrong with the data. For instance, if we take the changeset https://gerrit.onap.org/r/c/oom/+/85042 as an example, we can see that it's compose of 4 patcheset:
- patchset 1 -> 3 approvals
- patchset 2 -> 1 approval
- patchset 3 -> 2 approvals
- patchset 4 -> 1 approval

However in the data retrieved, the patchset 4 has one more approval, which is the last one of patchset 3. Since the last approval of patchset 3 is more recent that the approval for patchset 4, we end up with a negative time to first review. See the code reviews of `ikram` in the snapshot below.

captura_46

Perceval seems to work fine, so I'm going to update the code to discard the patchset reviews more recent than the patchset creation.

> Although most reviews were made by a bot, we shouldn't show negative numbers. The question would be whether or not to show time to first review when the patchset was rejected (-2) by a bot.

This requires some additional changes, because we should include SH information before assessing the time to first review

alpgarcia (approved):
Tested and working! Now reviews done before a patchset is sent are ignored for calculating the time of that patchset to its first review. There will be patchsets without time to first review as sometimes there are no reviews between patchsets. That's makes me wonder whether time to first review for patchsets makes sense at all.

Anyway, as the feature was requested as it is, I see this implementation as the correct approach to achieve the desired goal :)

PR #677, opened 2019-08-09
[dependencies] Update pandas dependency to 0.22.0

valeriocos:
This change is needed to align the dep version of pandas to the one used in SortingHat. It address https://github.com/chaoss/grimoirelab-sortinghat/issues/209


sduenas (approved):
LGTM

PR #680, opened 2019-08-19
[enriched-mediawiki] Remove identity/project info in get_rich_item

valeriocos:
This code removes the calls to add indentity and project information from the method `get_rich_item`. This code has no effect since it was already added in `get_rich_item_reviews`.


sduenas (approved):
LGTM

PR #682, opened 2019-08-29
[enriched-git] Remove credentials from Git URLs

valeriocos:
This code removes possible credentials declared on https URLs for Git repositories. This is the case of private repos on GitHub, which may be accessed by passing the username and token on the URL.

Tests have been provided accordingly.


sduenas (approved):
LGTM

PR #684, opened 2019-08-31
[ELK] Remove reference to ocean-unique-id info

valeriocos:
This PR removes the reference to ocean-unique-id from ELK. This attribute was used before the introduction of the attribute `uuid` in Perceval items. Tests and the schemas below have been updated, furthermore the method `fix_item` from the raw connector of Askbot has been removed.

- askbot
- bugzilla
- confluence
- discourse
- git
- github_issues
- irc
- jira
- kafka
- maniphest
- mbox
- mediawiki
- meetup
- redmine
- rss
- stackoverflow


alpgarcia (approved):
LGTM, thanks @valeriocos!

PR #687, opened 2019-09-11
[enriched-aoc] Include origin in AOC items

valeriocos:
This code includes the `origin` in the items generated by the AOC study. This change is needed to make the incrementality work in ceres, since it relies on this attribute (https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/ceres_base.py#L261)


sduenas (approved):
LGTM

PR #688, opened 2019-09-11
[enriched-gerrit] Handle missing user information

valeriocos:
This code handles missing user information (approvers, patchset and changeset authors) when calculating the time to first review for patchsets and changesets as well as the reviews' last status value.
In case the user information is missing, the aformentioned values are now calculated (before they were set to None).


sduenas (approved):
LGTM

PR #689, opened 2019-09-11
[enriched-onion] Filter bots from onion study

valeriocos:
This code filters the authors marked as bot, thus removing noise from the onion analysis.


alpgarcia (approved):
LGTM, thanks @valeriocos!

PR #690, opened 2019-09-12
fix num row error that GitHub complained about

GeorgLink:
GitHub noticed that there was one line that had more columns than the other lines.

This PR suggests to remove the duplicate "true" value. I hope that's the right way to fix this.


valeriocos (approved):
lgtm

PR #694, opened 2019-09-30
Add graal to setup.py, which is now a dependency.

jgbarah:


valeriocos (approved):
LGTM, thanks @jgbarah

Issue #701, opened 2019-10-08
Error enriching Discourse data when display name is disabled


jsmanrique:
It seems it is possible to disable display names in Discourse forums and it that case enrichment process raises an exception and fails:
```
mordred_1        | 2019-10-08 18:21:52,925 - grimoire_elk.elk - ERROR - Error enriching ocean from discourse (https://forum.mautic.org): 'display_username'
mordred_1        | Traceback (most recent call last):
mordred_1        |   File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/elk.py", line 663, in enrich_backend
mordred_1        |     total_ids = load_identities(ocean_backend, enrich_backend)
mordred_1        |   File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/elk.py", line 415, in load_identities
mordred_1        |     for identity in identities:
mordred_1        |   File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/enriched/discourse.py", line 78, in get_identities
mordred_1        |     user = self.get_sh_identity(post)
mordred_1        |   File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/enriched/discourse.py", line 89, in get_sh_identity
mordred_1        |     identity['name'] = post['display_username']
mordred_1        | KeyError: 'display_username'
```

How to reproduce it? Run an analysis on https://forum.mautic.org/

Wouldn't it be nice to manage such exception to make Discourse support functional even if such field is not present?

RCheesley:
Indeed, there is a system-wide option to disable the name field, which we use because many had erroneously entered an email address in an old system, and most users preferred to have username, not name.

I managed to work around this by simply changing the reference to name, to match what was used in username, but it would be more elegant if there was a way to check if the name field was in use, or have some way to deal with the situation.

valeriocos:
thank you @jsmanrique and @RCheesley for the issue. The fix will be in the next release (hopefully) :)

valeriocos:
Fixed with #702 . It will be available in the next release.

PR #702, opened 2019-10-09
[enriched-discourse] Handle missing author information

valeriocos:
This code allows to handle missing author information such as `display_username` and `username`, which are set to None.

Related to https://github.com/chaoss/grimoirelab-elk/issues/701


zhquan (approved):
LGTM

PR #703, opened 2019-10-10
[raw-jenkins] Handle multiple params on repo url

valeriocos:
This code allows to handle multiple parameters passed via the repo url, defined within the projects.json. This change is needed since the jenkins raw connector is redefining the method `get_p2o_params_from_url`, which contains the logic to handle `--filter`-like parameters.


zhquan:
LGTM

zhquan (approved):
LGTM

PR #705, opened 2019-10-15
[enriched-utils] Change `min_enrich` value

valeriocos:
This code changes how the value of `min_enrich` is calculated. In particular, the minimum value is based by preventing to set to zero seconds and milliseconds. This change is needed to avoid enriching over and over the last Git items.


alpgarcia (approved):
LGTM

PR #707, opened 2019-10-16
[elk] Remove filter raw prefix

valeriocos:
This code removes the support for filter-raw-prefix, which is a feature used to enrich items based on
the prefix of a given attribute value. This change aims at cleaning up the code with old feature not used
anymore.


valeriocos:
The CI is failing since a test is fetching data from a real remote server, which is under maintenance. As soon as the server is up again, the idea is to mock the call with the expected response.

zhquan (approved):
LGTM

PR #710, opened 2019-10-18
[discourse] Fix test and remove dead code

valeriocos:
This PR fixes the tests concerning the enrichment phases for discourse items by replacing the current origin (which isn't accessible) with a new one. Furthermore, the method `__show_categories_tree` is removed since not reachable.


zhquan (approved):
LGTM

PR #713, opened 2019-10-21
[Studies] Add new study to enable enriching a target index with extra fields

amrohassaan:
This adds new study that enables enriching a given index with some extra fields. This can be very useful to enable users to further enhance/tag data with some extra info relevant to their organisations.


valeriocos:
Thank you @amrohassaan for the pull request!

PR #716, opened 2019-10-22
[ELK] Support ODFE

valeriocos:
This code provides some fixes to support ODFE 1.2.0/elasticsearch 7.2.0.
This PR should be reviewed together with https://github.com/chaoss/grimoirelab-sirmordred/pull/362

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,32 +1,124 @@
 language: python
 
-python:
-  - "3.5"
-  - "3.6"
-
 sudo: false
 
 services:
-  - elasticsearch
   - mysql
 
-before_install:
-  - curl -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-6.1.0.deb && sudo dpkg -i --force-confnew elasticsearch-6.1.0.deb && sudo service elasticsearch restart
-  - pip install flake8
-  - pip install coveralls
-  - pip install httpretty==0.8.6
-  - pip install -r "requirements.txt"
+jobs:
+  include:
+    - python:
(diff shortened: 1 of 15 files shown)


sduenas (commented):
LGTM but it will worth that other person also reviews this. I also wonder why mappings are deleted and not used anymore.

valeriocos:
The deletion at https://github.com/chaoss/grimoirelab-elk/pull/716/files#diff-f7f4f8e73e7f8b04b43608dfaf3712ffL321 is due to the fact that code was for ES 2 and 5 (which is not supported anymore).

The deletion at https://github.com/chaoss/grimoirelab-elk/pull/716/files#diff-ea21b6959f1e04a43327ff62568ebf7bL483 is due to the fact that we don't use custom analyzers. I double-checked with @alpgarcia and we didn't find any reference of custom analyzers in use.

sduenas:
> The deletion at https://github.com/chaoss/grimoirelab-elk/pull/716/files#diff-f7f4f8e73e7f8b04b43608dfaf3712ffL321 is due to the fact that code was for ES 2 and 5 (which is not supported anymore).
> 
> The deletion at https://github.com/chaoss/grimoirelab-elk/pull/716/files#diff-ea21b6959f1e04a43327ff62568ebf7bL483 is due to the fact that we don't use custom analyzers. I double-checked with @alpgarcia and we didn't find any reference of custom analyzers in use.

Ok! Should someone else check this PR?

valeriocos:
Maybe @jjmerchante or @jsmanrique

PR #719, opened 2019-10-25
[enriched-enrich] Fix conversion projects.json to map

valeriocos:
This code fixes the way the projects map is built by processing the repositories under the `unknown`
section at the very end. This change is needed to avoid assigning repositories to the generic
project `Main` when they are also declared in other projects.

This error appears with Python 3.5, but not with Python 3.6 and it is related to how the dictionary
keys are ordered.

### DIFF
--- a/grimoire_elk/enriched/enrich.py
+++ b/grimoire_elk/enriched/enrich.py
@@ -205,7 +205,16 @@ class Enrich(ElasticItems):
         """
         ds_repo_to_prj = {}
 
-        for project in json:
+        # Sent the unknown project to the end of the list.
+        # This change is needed to avoid assigning repositories to
+        # the `Main` project when they exist in the `unknown`
+        # section and in other sections too.
+        project_names = list(json.keys())
+        if UNKNOWN_PROJECT in json:
+            project_names.remove(UNKNOWN_PROJECT)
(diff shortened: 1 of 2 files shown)


grimoire_elk/enriched/enrich.py
```diff
@@ -205,7 +205,17 @@ def __convert_json_to_projects_map(self, json):
...
-        for project in json:
+        # Sent the unknown project to the end of the list.
+        # This change is needed to avoid assigning repositories to
+        # the `Main` project when they exist in the `unknown`
+        # section and in other sections too.
+        project_names = list(json.keys())
+        found = json.get(UNKNOWN_PROJECT, None)
+        if found:
+            project_names.remove(UNKNOWN_PROJECT)
+            project_names.append(UNKNOWN_PROJECT)
```
sduenas: I would have done:

```
if UKNOWN_PROJECT in json:
    project_names.remove(UNKNOWN_PROJECT)
    project_names.append(UNKNOWN_PROJECT)
```

sduenas (approved):
LGTM

grimoire_elk/enriched/enrich.py
valeriocos: you are right, I'll modify the PR, thanks

PR #724, opened 2019-11-07
[elastic_items] Remove white spaces from repo url with labels

valeriocos:
This code strips out white spaces of repo urls having the param `--labels` enabled. This fix is needed to correctly match the url when enriching raw data.


zhquan (approved):
LGTM

PR #732, opened 2019-11-15
[enriched-enrich] Set empty item_sh when sortinghat is not set

valeriocos:
This code returns an empty item_sh when sortinghat isn't set.
This change is needed to allow ELK working without Sortinghat.


jjmerchante:
I have tested it with mordred and works for me. Thank you!

PR #734, opened 2019-11-18
Improve logs

zhquan:
Change to 'format()' in the logs

### DIFF
--- a/grimoire_elk/raw/elastic.py
+++ b/grimoire_elk/raw/elastic.py
@@ -17,6 +17,7 @@
 #
 # Authors:
 #   Alvaro del Castillo San Felix <acs@bitergia.com>
+#   Quan Zhou <quan@bitergia.com>
 #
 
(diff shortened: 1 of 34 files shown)


grimoire_elk/raw/elastic.py
```diff
@@ -274,7 +282,9 @@ def _items_to_es(self, json_items):
         if len(json_items) == 0:
             return
 
-        logger.debug("Adding items to Ocean for %s (%i items)" % (self, len(json_items)))
+        logger.debug("[{}] Adding items to Ocean for {} ({} items)".format(
```
valeriocos: Could replace `ocean` with `raw`?

valeriocos (commented):
Overall it looks great, thanks @zhquan . I left a tiny comment to replace the term `ocean`.

valeriocos (approved):
LGTM

PR #737, opened 2019-11-19
[enriched-github2] New enricher to handle github comments

valeriocos:
This code proposes a new github enricher to handle github comments and emotions. The old enricher is left in place to avoid breaking changes with customer dashboards.

The new enricher is added to the list of connectors, it can be defined in the mordred setup.cfg as follows:
```
[github2]
api-token = ...
raw_index = github_chaoss_raw
enriched_index = github_chaoss_enriched
sleep-for-rate = true

[github2:prs]
api-token = ...
raw_index = github-prs_chaoss_raw
enriched_index = github-prs_chaoss_enriched
sleep-for-rate = true
category = pull_request
no-archive = true
```

Similar names must be present in the projects.json

Tests and schemas have been added accordingly.


alpgarcia (changes requested):
I've been reviewing Sigils panels and the new indexes seem to have at least the same info we had before (except painless fields that are created at index pattern level). 

I suggest the following changes:
* I'd modify the code to get fields `is_github_*` `is_github2_*` for future compatibility. Imho these fields should not depend on the enricher because if we use them in the dashboards we would need to modify them if the enricher name changes.
* `issue_id_in_repo` should probably in issues too. I'd say all `issue_*` fields should be always present to allow to filtering regardless item type. Same for Pull Requests equivalent fields.
* I wonder whether having `issue_closed_at` could be useful for some metric regarding comments. For instance number of comments of issues closed at a give date...
* I see no reactions information for issues/PRs, only for comments, but I can react to them.

valeriocos:
Comments addressed @alpgarcia , please check

alpgarcia (changes requested):
One last change. Reactions object should be flatten for consistency's sake. Now it is stored as follows:
```
"reactions": {
            "properties": {
              "+1": {
                "type": "long"
              },
              "-1": {
                "type": "long"
              },
              "confused": {
                "type": "long"
              },
              "eyes": {
                "type": "long"
              },
              "heart": {
                "type": "long"
              },
              "hooray": {
                "type": "long"
              },
              "laugh": {
                "type": "long"
              },
              "rocket": {
                "type": "long"
              },
              "total_count": {
                "type": "long"
              }
            }
},
```

alpgarcia (approved):

PR #738, opened 2019-11-25
[enriched-git] Handle missing local git repo

valeriocos:
This code allows to handle the Perceval RepositoryErrors triggered when the repository isn't found locally. The exception is caught and a warning message is logged.


zhquan (approved):
LGTM

PR #743, opened 2019-12-08
[elk] Replace ElasticConnect, ElasticWriter exceptions with ElasticError

valeriocos:
This code models the errors coming from the Elastic class by defining the exception `ElasticError`. This new error replaces the exceptions ElasticConnectException and ElasticWriteException.


zhquan (approved):
LGTM

PR #744, opened 2019-12-08
[utils] Add 429 error to status_force_list

valeriocos:
This code enables the retrying mechanism on 429 errors, when connecting to elasticsearch. This code is needed to prevent failures during the CI tests for OpenDistro


zhquan (approved):
LGTM

PR #749, opened 2019-12-23
[ELK] Support feelings GitHub2 data

valeriocos:
This PR addresses https://github.com/chaoss/grimoirelab/issues/258

This code includes the enrich_feelings study, which allows to add sentiment and emotion data to a target enriched index.

The study requires the URL of the NLP REST service, plus a list of attributes containing the text to
be processed. The logic of the study is described below.

All items in the target enriched index not including the fields `has_sentiment` or `has_emotion` are retrieved. For each item, the target attribute is identified, then the text is convert to plain text using the NLP REST endpoint `plainTextBugTrackerMarkdown`. The obtained text is processed to strip out code parts using the endpoint `code`, the text left is passed to the endpoints `sentiment` and `emotion`, which are used to derive sentiment and emotion labels for the input text. The labels collected are saved in the corresponding item on the attributes `feeling_sentiment` and `feeling_emotion`. Furthermore, also the attributes `has_sentiment = 1` and `has_emotion = 1` are added to the item to prevent the recalculation of feeling study for items already analysed.

This study is showcased on github2 data. The study can be configured with the setup.cfg of mordred in the following way:
```
[github2]
api-token = ...
raw_index = github_chaoss_crossminer
enriched_index = github_chaoss_crossminer_enriched
sleep-for-rate = true
no-archive = true
studies = [enrich_feelings]

[enrich_feelings]
attributes = [title, body]
nlp_rest_url = http://localhost:2901
```

Things to be done:
- [x] Add docstrings
- [x] Add tests
- [x] Check support for OpenDistro
- [x] Check results with EHU


zhquan (approved):
LGTM

Issue #759, opened 2020-01-13
[cocom] [colic] Dashboards display tokens in URLs


svdo:
This is the same issue as was fixed by #682, but now for the Code Complexity and the Code Licenses dashboards. Both the "Repositories" and the "All Files: Code complexity, LOC and ratios" panels on the Code Complexity dashboard, as well as the "Repositories" and "All Files: License & Copyright Definition" panels on the Code Licenses dashboard display username and token in the URLs.

valeriocos:
Fixed with #763 and chaoss/grimoirelab-sigils#414

Issue #755, opened 2019-12-31
[cocom] analysis fail with some repository


florentk:
I got this error for some on our repositories : 
```
2019-12-31 17:40:08 > 2019-12-31 16:40:05,843 - grimoire_elk.elk - ERROR - Error enriching ocean from cocom (https://github.com/kiwix/tools): 'message'
2019-12-31 17:40:08 > Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/elk.py", line 672, in enrich_backend
    enrich_count = enrich_items(ocean_backend, enrich_backend)
  File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/elk.py", line 452, in enrich_items
    total = enrich_backend.enrich_items(ocean_backend)
  File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/enriched/cocom.py", line 223, in enrich_items
    rich_items = self.get_rich_items(item)
  File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/enriched/cocom.py", line 183, in get_rich_items
    eitem['message'] = entry['message']
2019-12-31 17:40:08 > KeyError: 'message'
```

https://github.com/kiwix/metrics/issues/16

valeriocos:
Hi @florentk, sorry for the late reply!

If the repo is public, could you share it so I'll be able to replicate the issue?

florentk:
@valeriocos yes you cas try with this https://github.com/kiwix/tools

valeriocos:
The problem is due to a commit without message, see details below.
```
slimbook@slimbook-KATANA:/tmp/tools$ git show 899e6b21130d10811e0d9853194556e7cb4d247c
commit 899e6b21130d10811e0d9853194556e7cb4d247c
Author: coulonito <coulonito@users.sourceforge.net>
Date:   Thu Mar 1 09:34:44 2007 +0000

diff --git a/chrome/content/interfacewiki/default.css b/chrome/content/interfacewiki/default.css
new file mode 100644
index 0000000..525f345
--- /dev/null
+++ b/chrome/content/interfacewiki/default.css
@@ -0,0 +1,4 @@
+#wk-noresult {
+
+  color:#aaa;
+}
diff --git a/chrome/content/interfacewiki/img/logo.png b/chrome/content/interfacewiki/img/logo.png
new file mode 100644
index 0000000..840de7d
[...]

PR #763, opened 2020-01-14
graal-related enrichers

valeriocos:
This code addresses https://github.com/chaoss/grimoirelab-elk/issues/759 and https://github.com/chaoss/grimoirelab-elk/issues/755. Thus, it allows to handle commits without message in colic and cocom enrichers. Furthermore, it stores the origin without credentials in a separated attributes, which is later on visualized in the dashboards.

Note that the origin and the credentials in it can be visualized by inspecting the raw or/and enriched index.

Tests and schemas have been updated accordingly.

Related to https://github.com/chaoss/grimoirelab-sigils/pull/414


sduenas (changes requested):
The code looks ok but would you mind to split this PR in two pices? One resolving each bug.

sduenas (approved):
LGTM

PR #766, opened 2020-01-15
Add support for releasing scroll and waiting for available scroll, fixes #758

### LINKED ISSUE #758: [Bug?] ES 7.X exceeding maximum scrolls [500] is ignored. ELK tools should remove scroll when they no more need it.
So when I run a big batch of `p2o.py` commands (about 1900) using about 32 in parallel (running on 2 kubernetes nodes each having 16 CPU cores, so 32 `p2o.py` processes at the same time) I'm hitting ES 7.X error "too many scrolls open".

The funny part is that I would never detect that because `p2o.py` runs "as usual" reporting no failure and even no warning, the result is that no new data comes into the enriched index.

This runs OK on ES 6.X probably because the number of allowed scrolls is bigger (both 6.X and 7.X) initially have no setting to be checked by `curl (...)/_cluster/settings?pretty" | grep max_open_scroll`. You can set that value via specific `curl -XPUT (...)/_cluster/settings` call. My guess is that 6.x has no limit for that value or the limit is much higher than 500 (I was monitoring this value via `curl (...)/_nodes/stats/indices/search?pretty | grep "open_contexts"` while sync was running and 6.X allowed much over 1000 while 7.X go to 500 and stopped.

I've detected this just by pure luck, my other script was trying to save entire index data (one index) into JSON file using scroll API and it failed with
```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "exception",
        "reason" : "Trying to create too many scroll contexts. Must be less than or equal to: [500]. This limit can be set by changing the [search.max_open_scroll_context] setting."
      }
    ],
    "type" : "search_phase_execution_exception",
    "reason" : "all shards failed",
    "phase" : "query",
    "grouped" : true,
    "failed_shards" : [
      {
        "shard" : 0,
        "index" : "sds-hyperledger-fabric-github-issue",
        "node" : "p_8cDIekSPWjJ4mEitZ2KA",
        "reason" : {
          "type" : "exception",
          "reason" : "Trying to create too many scroll contexts. Must be less than or equal to: [500]. This limit can be set by changing the [search.max_open_scroll_context] setting."
        }
      }
    ]
  },
  "status" : 500
}
```

So I've run one of `p2o.py` commands manually while. 7.X was on the 500 scrolls limit and it just finished OK, without any warning or error.

==========================

So the problem is:

1) why `p2o.py` creates more than 32 scrolls when only 32 `p2o.py` instances are run in parallel, no more? My guess is that grimoire-elk library (which is used by `p2o.py`) is *NOT* clearing index after it no more uses it, see:
https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/elastic_items.py#L201-L205

The scroll is set to even 10 minutes there (according to comment) but I don't see scroll released after no more used. ElasticSearch API docs strongly suggest that scrolls should be removed after they're no more needed, see: https://www.elastic.co/guide/en/elasticsearch/reference/7.5/search-request-body.html#_clear_scroll_api

My guess is that having > 1000 scrolls are just leftovers from finished `p2o.py` calls. I think each call should clean up its scroll, like for example here: https://github.com/LF-Engineering/darst/blob/master/utils/get_all_index_docs.sh#L69-L70

2) I think `p2o.py` and other stuff using https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/elastic_items.py file should check if scrolls are available and if not it should return an error or at least a warning. The other option is to wait for some scroll(s) to become available. But the most important is to FREE the one that was used in the current instance and is no more needed!

I think (not yet 100% sure) that this is the real cause of https://github.com/chaoss/grimoirelab-elk/issues/757

cc @sduenas @valeriocos

lukaszgryglicki:
Wait a configurable amount of time (in seconds) via `--scrol=wait`, for example `--scroll-wait 30` (default is 900)
Fixes #758 
Signed-off-by: Lukasz Gryglicki <lukaszgryglicki@o2.pl>

### DIFF
--- a/grimoire_elk/elastic_items.py
+++ b/grimoire_elk/elastic_items.py
@@ -182,7 +182,7 @@ class ElasticItems:
                 time.sleep(1)
                 sec -= 1
                 page = self.get_elastic_items(scroll_id, _filter=_filter, ignore_incremental=ignore_incremental)
-                if not 'too_many_scrolls' in page:
+                if 'too_many_scrolls' not in page:
                     logger.debug("Scroll acquired after {} seconds".format(self.scroll_wait - sec))
                     break
 
@@ -324,7 +324,7 @@ class ElasticItems:
         try:
             res = self.requests.post(url, data=query_data, headers=headers)
             if self.too_many_scrolls(res):
-                return { 'too_many_scrolls': True }
+                return {'too_many_scrolls': True}
             res.raise_for_status()
             rjson = res.json()
         except Exception:
@@ -345,4 +345,4 @@ class ElasticItems:
             and len(r['error']['root_cause']) > 0
             and 'reason' in r['error']['root_cause'][0]
             and 'Trying to create too many scroll contexts' in r['error']['root_cause'][0]['reason']
-            )
+        )
--- a/grimoire_elk/elastic_items.py
+++ b/grimoire_elk/elastic_items.py
@@ -182,6 +182,9 @@ class ElasticItems:
                 time.sleep(1)
                 sec -= 1
                 page = self.get_elastic_items(scroll_id, _filter=_filter, ignore_incremental=ignore_incremental)
+                if not page:
+                    logger.debug("Waiting for scroll terminated")
+                    break
                 if 'too_many_scrolls' not in page:
                     logger.debug("Scroll acquired after {} seconds".format(self.scroll_wait - sec))
                     break
--- a/grimoire_elk/elastic_items.py
+++ b/grimoire_elk/elastic_items.py
@@ -25,6 +25,7 @@
 import json
 import logging
 import re
+import time
 
 from .enriched.utils import get_repository_filter, grimoire_con
 from .elastic_mapping import Mapping
@@ -46,6 +47,7 @@ class ElasticItems:
     # In large projects like Eclipse commits, 100 is too much
     # Change it from p2o command line or mordred config
     scroll_size = 100
+    scroll_wait = 900
 
     def __init__(self, perceval_backend, from_date=None, insecure=True, offset=None):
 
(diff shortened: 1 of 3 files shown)


lukaszgryglicki:
cc @valeriocos @jgbarah @sduenas

valeriocos:
Sorry for the late reaction @lukaszgryglicki ! Do you know how I can test your PR? How many instances of p2o should I launch to see the error popping out?

lukaszgryglicki:
Start ES server, must be >= 7.X (major version): `docker run -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:7.5.1`

Then set scrolls to some low number, for example, 2. Then start > 2 p2o.py commands (for example 4) in parallel (ideally some long ones). 

The new version will start and finish 2 of them while 3rd will wait (see that with `--debug` or `-g` flag). Once 1st or 2nd finishes 3rd will continue, then 4th (only 2 at a time). You can specify max wait for scroll via `--scroll-wait 300` (5 minutes), default is 900 (15 minutes).

The old version will enrich only 2 of 4 and silently ignore errors with 3rd and 4th failing to use the scroll.

You can also monitor the number of scrolls used during p2o.py is running.

The new version will use max 2 and will drop to 0 after all p2o finished.

The old version will also use 2 (due to maximum set) but will not free them after finishing (they will stay for 5 minutes as this is the scroll setting used by p2o). So when you drop indexes created by old p2o and try old version again it will create raw indexes OK but enrich nothing (in those 5 minutes, because the previous run has not freed them) and report no error.

To set max scrolls you can use this script: https://github.com/LF-Engineering/darst/blob/master/utils/set_max_scrolls.sh
To check how many scrolls are currently in use: `curl [[es_url]]/_nodes/stats/indices/search?pretty 2>/dev/null | grep "open_contexts"`.

In case of ES running from docker as I suggested [[es_url]] will be http://localhost:9200

valeriocos (approved):
LGTM @lukaszgryglicki , sorry for the late review.

lukaszgryglicki:
Thanks!

PR #769, opened 2020-01-21
[enriched-enrich] Add geolocation study

valeriocos:
This study includes geo points information (latitude and longitude) based on the value of the `location_field`. The coordinates are retrieved using Nominatim through the geopy package (which required to update the setup.py and requirements.txt), and saved in the `geolocation_field`.

All locations included in the `location_field` are retrieved from
the enriched index. For each location, the geo points are retrieved and saved
to the `geolocation_field`. In case the geo points are not retrieved,
the location is ignored.

The example below shows how to activate the study by modifying
the setup.cfg. The study `enrich_geolocation:user` retrieves location data
from `user_location` and stores the geo points to `user_geolocation`. In a
similar manner, `enrich_geolocation:assignee` takes the data from
`assignee_location` and stores the geo points to `assignee_geolocation`.

```
[github]
raw_index = github_issues_chaoss
enriched_index = github_issues_chaoss_enriched
api-token = ...
sleep-for-rate = true
...
studies = [enrich_geolocation:user, enrich_geolocation:assignee]

[enrich_geolocation:user]
location_field = user_location
geolocation_field = user_geolocation

[enrich_geolocation:assignee]
location_field = assignee_location
geolocation_field = assignee_geolocation
```

The study is enabled for github and github2 enrichers. Tests have been added accordingly.

This PR removes also some dead code found in the github enricher (https://github.com/chaoss/grimoirelab-elk/commit/e810af9b6a5f56465a78916226f0ee2c4956c335)

Related to https://github.com/chaoss/grimoirelab-sirmordred/pull/386

### DIFF
--- a/grimoire_elk/enriched/enrich.py
+++ b/grimoire_elk/enriched/enrich.py
@@ -31,6 +31,7 @@ import pkg_resources
 from functools import lru_cache
 
 from elasticsearch import Elasticsearch, RequestsHttpConnection
+from geopy.geocoders import Nominatim
 
 from perceval.backend import find_signature_parameters
 from grimoirelab_toolkit.datetime import (datetime_utcnow, str_to_datetime)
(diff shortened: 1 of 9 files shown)


alpgarcia (changes requested):
After talking to @valcos and reading [Nominatim use rules](https://operations.osmfoundation.org/policies/nominatim/) I think we should implement a cache in our side, specifically due to avoid the risk of falling within:
```
Bulk Geocoding
As a general rule, bulk geocoding of larger amounts of data is not encouraged. If you have regular geocoding tasks, please, look into alternatives below. Smaller one-time bulk tasks may be permissible, if these additional rules are followed

- limit your requests to a single thread
- limited to 1 machine only, no distributed scripts (including multiple Amazon EC2 instances or similar)
- Results must be cached on your side. Clients sending repeatedly the same query may be classified as faulty and blocked.
```

I would propose to use the index itself as cache. We can query for `<location string>:<value> AND <geolocation_field>:*` and just get one item with only the geolocation fields we need. That should be a light query and would avoid unneeded queries to Nominatim. 

Even lighter would be having a separate smaller dedicated cache index that would also allow us to reuse the same data among different instances without querying the server.

valeriocos:
PR modified according to your comment @alpgarcia. The cache solution works fine also to restore the geolocation information of the last items for each origin which are constantly overwritten.

zhquan (approved):
LGTM

PR #770, opened 2020-01-23
Let pandas be >= 0.22.0

jgbarah:
If we pin pandas at 0.22.0, it doesn't build
with Python 3.7.x, apparently because of problems
when building numpy for pandas.

Letting pandas be greater or equal than 0.22.0
doesn't seem to break anything, since platforms
having only 0.22.0, as Travis had in July'19,
would still fulfill the dependency.

Signed-off-by: Jesus M. Gonzalez-Barahona <jgb@gsyc.es>


valeriocos (approved):
lgtm, thanks

PR #771, opened 2020-01-24
[elastic] Enable loading of complex aliases

valeriocos:
This code allows to load aliases containing filters, which at the moment are manually loaded.


zhquan (approved):
LGTM

PR #773, opened 2020-01-25
[git-enriched] Remove github commit support

valeriocos:
This code removes the support to include author and committer information extracted from github commits, when enriching git data. This feature isn't used, in its current state it won't work for github enterprise instances and it may slow down the enrichment of git data.

Related code present in elk.py and sortinghat_gelk.py has been removed accordingly.

Review together with: https://github.com/chaoss/grimoirelab-sirmordred/pull/390
Deprecates https://github.com/chaoss/grimoirelab-elk/pull/753


zhquan (approved):
LGTM

Issue #767, opened 2020-01-16
[dockerhub] backend enrichment error (python exception)


lukaszgryglicki:
Python exception:

`p2o.py --enrich --index raw --index-enrich rich -e [redacted_es_7_url] -g --db-host mariadb.mariadb --db-sortinghat sortinghat --db-user sortinghat --db-password [redacted] dockerhub hyperledger sawtooth-next-directory`:

```
2020-01-16 13:48:53,765 Error enriching raw from dockerhub (https://hub.docker.com/hyperledger/sawtooth-next-directory): unorderable types: NoneType() <= NoneType()
Traceback (most recent call last):
  File "/repos/grimoirelab-elk/grimoire_elk/elk.py", line 666, in enrich_backend
    enrich_count = enrich_items(ocean_backend, enrich_backend)
  File "/repos/grimoirelab-elk/grimoire_elk/elk.py", line 452, in enrich_items
    total = enrich_backend.enrich_items(ocean_backend)
  File "/repos/grimoirelab-elk/grimoire_elk/enriched/dockerhub.py", line 166, in enrich_items
    if image_date <= rich_item['last_updated']:
TypeError: unorderable types: NoneType() <= NoneType()
2020-01-16 13:48:53,766 [dockerhub] Done enrichment for https://hub.docker.com/hyperledger/sawtooth-next-directory
2020-01-16 13:48:53,767 Enrich backend completed
2020-01-16 13:48:53,767 Finished in 0.03 min
```

valeriocos:
Fixed with #774

lukaszgryglicki:
Thanks!

PR #774, opened 2020-01-27
[enriched-dockerhub] Handle never updated images

valeriocos:
This PR addresses https://github.com/chaoss/grimoirelab-elk/issues/767, thus it allows to ignore dockerhub images, which don't have a last updated value. This is the case of https://hub.docker.com/r/hyperledger/sawtooth-next-directory, that doesn't have any activity.


zhquan (approved):
LGTM

PR #778, opened 2020-02-04
[raw-bugzillarest] Prevent indexing of clones attribute

valeriocos:
This code prevents the indexing of the attribute clones, which isn't used in the enriched data.

Tests have been modified accordingly.


zhquan (approved):
LGTM

Issue #779, opened 2020-02-05
Error enriching ocean from telegram / Error feeding raw from telegram


antonionardella:
Hello,

I am using the grimoirelab/full docker container.
This is the backend configuration:

projects.json
```
        "telegram": [
            "[REDACTED BOT NAME]"
        ]
```

credentials.cfg
``` 
[telegram]
api-token = [REDACTED BOT API TOKEN]
raw_index = telegram_iota_raw
enriched_index = telegram_iota_enriched

```

I get these errors in the logs:

* username

```
grimoire       | 2020-02-05 15:48:34,978 - grimoire_elk.elk - ERROR - Error enriching ocean from telegram (https://telegram.org/[REDACTED]): 'username'
grimoire       | Traceback (most recent call last):
grimoire       |   File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/elk.py", line 657, in enrich_backend
grimoire       |     total_ids = load_identities(ocean_backend, enrich_backend)
grimoire       |   File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/elk.py", line 415, in load_identities
grimoire       |     for identity in identities:
grimoire       |   File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/enriched/telegram.py", line 80, in get_identities
grimoire       |     identity = self.get_sh_identity(message['from'])
grimoire       |   File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/enriched/telegram.py", line 69, in get_sh_identity
grimoire       |     identity['username'] = from_['username']
grimoire       | KeyError: 'username'
```

* message_id
```
grimoire       | 2020-02-05 15:48:24,612 - grimoire_elk.elk - ERROR - Error feeding raw from telegram (https://telegram.org/[REDACTED]): 'message'
grimoire       | Traceback (most recent call last):
grimoire       |   File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/elk.py", line 228, in feed_backend
grimoire       |     ocean_backend.feed(**params)
grimoire       |   File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/raw/elastic.py", line 230, in feed
grimoire       |     self.feed_items(items)
grimoire       |   File "/usr/local/lib/python3.5/dist-packages/grimoire_elk/raw/elastic.py", line 246, in feed_items
grimoire       |     for item in items:
[...]

PR #781, opened 2020-02-08
[enriched-telegram] Handle missing username in Telegram messages

valeriocos:
This PR addresses part of https://github.com/chaoss/grimoirelab-elk/issues/779, thus it handles Telegram messages without an username.


zhquan (approved):
LGTM

PR #785, opened 2020-02-17
[enriched-jira] Handle missing creator/ assignee displayName and name

valeriocos:
This code allows to handle creators and assignees without displayName or name.

Tests data have been modified accordingly.


zhquan (approved):
LGTM

Issue #761, opened 2020-01-14
[GitHub] Add study to create a "backlog" index


florentk:
With the current Github enriched datas, we have the backlog only from actual data, we have not the evolution.  For example we have not the evolution of the number of opened issues or the evolution of average issue resolution time. We can compute these metrics in add a new study index based on Github enirched datas.

We have implement this in a [fork of grimoirelab_elk](https://github.com/florentk/grimoirelab-elk/tree/kiwix-github-evolution-backlog) for the dev of [Kiwix metric](http://metrics.kiwix.org). Are you interested in including it as a pull request ?

jgbarah:
Thanks a lot, this is quite interesting!

For this kind of stuff, we usually write an specific "study", which is a way of producing a "special" index (such as the one we write for the "areas of code" panels). This way, we keep the core of GrimoireELK reasonably simple, while we can have those optional studies performed as a part of what SirMordred does.

@valeriocos, @sduenas do you agree this would be the best approach to integrate this code? Do you have any pointer to studies, so that @florentk can consider if it makes sense in this case?

alpgarcia:
Related to https://github.com/chaoss/grimoirelab-sigils/issues/410

valeriocos:
Thank you @florentk for sharing your work. I'll have a look at it tomorrow.

florentk:
@valeriocos Do you have look my proposition ? Can I change something to integrate my code in a branche of your repository (is the best way for us !)

valeriocos:
Hi @florentk sorry for the late reply! Yes, I had a look to your fork and the approach looks ok!
Could you submit a PR with the study? There, we can comment on some details of the study, WDYT?

Thanks and sorry again!

valeriocos:
This issue has been solved with:
- https://github.com/chaoss/grimoirelab-elk/pull/789
- https://github.com/chaoss/grimoirelab-elk/pull/790
- https://github.com/chaoss/grimoirelab-sirmordred/pull/408

PR #789, opened 2020-02-24
[github] GitHub evolution backlog

florentk:
This PR fix the issue https://github.com/chaoss/grimoirelab-elk/issues/761 to add additional index to compute
the chronological evolution of opened issues and average opened time 
issues.

For each repository and label, we start the study on repository
creation date until today with a day interval (default). For each date
we retrieve the number of open issue at this date by difference between
number of opened issues and number of closed issues. In addition, we
compute the average opened time for all issues open at this date.

To differenciate by label, we compute evolution for bugs and all others
labels (like "enhancement","good first issue" ... ), we call this
"reduced labels". We need to use theses reduced labels because the
complixity to compute evolution for each combinaison of labels would be
too big. In addition, we rename "bug" label to "bugs" in reduced labels.

### DIFF
--- a/grimoire_elk/enriched/github.py
+++ b/grimoire_elk/enriched/github.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2015-2019 Bitergia
+# Copyright (C) 2015-2020 Bitergia
 #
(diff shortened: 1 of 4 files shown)


valeriocos (approved):
LGTM, thanks @florentk :)

PR #790, opened 2020-02-24
[git] Add "tags" field in git enriched index 

florentk:
add "tags" field in git enriched index to get the number of releases.


valeriocos (approved):
LGTM, thanks @florentk

PR #791, opened 2020-02-25
[github2] Fix loading identities

valeriocos:
This code fixes the way the identities are loaded in the method `get_identities` to make the process
more efficient. The problem is that the loop that loads the identities found in comments is executed
for each identity type, however this loop should be executed just once.

The fix allows to execute the aforementioned just once.


jjmerchante:
LGTM!

PR #792, opened 2020-02-25
[elk] Remove support for arthur

valeriocos:
This code removes the references, methods depedencies and config params used to fetch data from arthur. This change is needed since the integration with arthur will be implemented in a different way (as discussed in the roadmap).

Related to chaoss/grimoirelab#299


vchrombie:
Hi @valeriocos 

I just saw the issue in the grimoirelab repository about the removal of support for KingArthur for now. I just had a small doubt.

Is "removing the support" means, do we need to remove configuration files too?

I asked this because I just had a look at your `valeriocos:issue-gl-299` branch and the docker configurations and the scripts in the utils folder are still present. I just want to know if you purposefully let it go, for future use, as it was planned to change the way KingArthur would be used.

valeriocos:
Hi @vchrombie, the docker configurations and scripts are old/not used (except for p2o) and I guess they could be removed, but I want to ask some colleagues before proceeding. In case this is possible, we can target their removal in a different PR.

vchrombie:
Okay. Thanks for the reply :slightly_smiling_face: 

> we can target their removal in a different PR, and you can take care of it., WDYT?

Sure, I can help you with it. Please let me know after you discuss with your colleagues.

valeriocos:
sure! thanks @vchrombie

zhquan (approved):
LGTM

PR #793, opened 2020-02-28
[doc] Update readme for secure elasticsearch

snack0verflow:
Added some instruction for secure elasticsearch and docker users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,8 +120,8 @@ Tests are located in the folder [tests](https://github.com/chaoss/grimoirelab-el
 - the [dependencies](https://github.com/chaoss/grimoirelab-elk/blob/master/requirements.txt) installed or available (the latter applies to the dependencies with the other GrimoireLab repositories).
 
 Then you need to:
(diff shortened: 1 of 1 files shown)


valeriocos (commented):
Thank you for the PR @snack0verflow !
Quick question, why the example in the third entry (populate the database ..) is different from the example in the second entry (create the databases ..)? It wouldn't be easy to have something like the examples below or I'm missing something?
- `docker exec -it <container id> mysql -u root -e "create database test_sh"` (for the second entry)
- `docker exec -it <container id> mysql -u root test_projects < tests/test_projects.sql"` (for the third entry)

snack0verflow:
@valeriocos Thanks for pointing this out, the -t option is for TTY input which doesn't work while populating the database. I checked again and I think we can get rid of -t from both the cases.

valeriocos:
thanks @snack0verflow for the quick reply!

snack0verflow:
@valeriocos Why did the coverage decrease?

valeriocos:
is your branch in sync with the master one?

valeriocos (approved):
LGTM, thanks @snack0verflow

PR #801, opened 2020-03-10
[grimoire_elk] Add support for multiple affiliations

valeriocos:
This code allows to store in the enriched indexes information of multiple affiliations. Such a data is stored:
- as a list in the sortinghat-related attribute `<rol>_multi_org_names`
  Ex.: author_multi_org_names
- as single sortinghat-related attributes `<rol>_multi_org_name_<num>`
  Ex.: author_multi_org_name_0, author_mult_org_name_1

The code that deals with the identities refresh has been modified as well (https://github.com/chaoss/grimoirelab-elk/commit/20ea441550cb200b5f811b25940bbe7cc20f4de6#diff-a94b51d8c8a4c2e33e52cff8968a0b3c) to make sure that the affiliations stored in the list `<rol>_multi_org_names` is aligned with the attributes `<rol>_multi_org_name_<num>`. The change consists of removing the attributes `<rol>_multi_org_name_<num>` and add them again. It was need to cover the cases when the number of affiliations decreases. In this scenario, the attributes `<rol>_multi_org_name_<num>` with `<num`> greater than the new list size `<rol>_multi_org_names` were not modified/updated, thus introducing a misalignment between the list and single attributes.

Tests have been updated accordingly. Special cases are Jira and Gerrit, which have been separated in different commits (to ease the review).

Schemas have not been updated.
Once the PR is accepted, the index patterns in Sigils should be updated.

Related to: https://github.com/chaoss/grimoirelab-cereslib/pull/28

### DIFF
--- a/grimoire_elk/elk.py
+++ b/grimoire_elk/elk.py
@@ -21,6 +21,7 @@
 
 import inspect
 import logging
+import re
 
 from elasticsearch import Elasticsearch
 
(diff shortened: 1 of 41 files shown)


valeriocos:
In order to store the minimal info in the enriched indexes, the single sortinghat-related attributes <rol>_multi_org_name_<num> (e.g., author_multi_org_name_0) could be removed. They could be automatically derived using scripted fields.

PR #806, opened 2020-03-16
[raw-gerrit] Handle long text for data.patchSets.comments.message

valeriocos:
This code allows to handle immense term exceptions due to the lenght of the attribute `data.patchSets.comments.message`.

Tests have been modified accordingly.


zhquan (approved):
LGTM

PR #809, opened 2020-03-17
Updated test_git.py

kshitij3199:
Updated test_raw_to_enrich_projects method


valeriocos (approved):
LGTM, thanks @kshitij3199 . Please consider to add the same test to the other connectors, thanks :)

PR #810, opened 2020-03-18
added warning for trailing slashes in URL

imnitishng:
Encountered an error where the github backend would consider owner of the repository as its name and repository argument was left empty because of an extra forward slash (/) in the URL of repositories in `projects.json` (which is pretty common while copying and pasting a URL). Thought of adding this warning if this situation occurs.

### DIFF
--- a/grimoire_elk/raw/github.py
+++ b/grimoire_elk/raw/github.py
@@ -21,7 +21,9 @@
 
 from .elastic import ElasticOcean
 from ..elastic_mapping import Mapping as BaseMapping
+import logging
 
+logger = logging.getLogger(__name__)
 
 class Mapping(BaseMapping):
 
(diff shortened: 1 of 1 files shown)


valeriocos:
Thank you @imnitishng , I'm not sure about this fix. I'm wondering if:
- the validation of the input should be done far before reaching the raw/github.py
- there is no fix needed. If a user feeds something wrong to the system, a error should be thrown.

Any thoughts?

imnitishng:
Okay! Travis is also failing so I will try looking into the error and perform a check before input reaches this far. Thank You!

valeriocos:
Hi @imnitishng, can we close this pull request?

imnitishng:
Yes sure.
I did not understand the toolchain well enough when I opened this. Now I
realise this was not a need.
Thank you for you support all along.

valeriocos:
Thank you @imnitishng !

Issue #822, opened 2020-03-26
Anonymize user information coming from Perceval


jjmerchante:
We need a way to anonymize the personal information that comes from Perceval. Names, emails and information that may be related to the users of each backend must be hidden.

The information stored in the Elasticsearch indexes should only be identifiers that cannot be related directly with the user.

I will open a pull request for Git, GitHub, Gitlab and Meetup that are the current data sources of Cauldron.

valeriocos:
Fixed with #824

PR #823, opened 2020-03-26
Make anonymization optional for each backend

jjmerchante:
We need a way to anonymize the personal information that comes from Perceval: names, emails and information that may be related to the users of each backend must be hidden.

The information stored in the Elasticsearch indexes should only be identifiers that cannot be related directly with the user.

In this merge request, information related with the users is hashed or removed.

You can set the parameter `anonymize` to true in the sirmordred configuration:
```
[git]
...
anonymize = true

[github]
...
filter-classified = true
anonymize = true

[gitlab]
...
anonymize = true

[meetup]
...
anonymize = true
```

Related with https://github.com/chaoss/grimoirelab-elk/issues/822

Depends on https://github.com/chaoss/grimoirelab-elk/pull/823

### DIFF
--- a/grimoire_elk/elk.py
+++ b/grimoire_elk/elk.py
@@ -48,7 +48,8 @@ requests_ses = grimoire_con()
 
 def feed_backend(url, clean, fetch_archive, backend_name, backend_params,
                  es_index=None, es_index_enrich=None, project=None,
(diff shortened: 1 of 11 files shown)


valeriocos:
Closed in favor of #824

PR #824, opened 2020-03-28
Make anonymization optional for each backend in a new module

jjmerchante:
This pull request is a different implementation for the issue https://github.com/chaoss/grimoirelab-elk/issues/822
First pull request was https://github.com/chaoss/grimoirelab-elk/pull/823

This PR depends on sirmordred PR https://github.com/chaoss/grimoirelab-sirmordred/pull/442

Description:

We need a way to anonymize the personal information that comes from Perceval: names, emails and information that may be related to the users of each backend must be hidden.

The information stored in the Elasticsearch indexes should only be identifiers that cannot be related directly with the user.

In this merge request, information related with the users is hashed or removed.

You can set the parameter `anonymize` to true in the sirmordred configuration:
```
[git]
...
anonymize = true

[github]
...
filter-classified = true
anonymize = true

[gitlab]
...
anonymize = true

[meetup]
...
anonymize = true
```


grimoire_elk/identities/git.py
```diff
@@ -0,0 +1,47 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
```
valeriocos: Please add your name as author

grimoire_elk/identities/git.py
```diff
@@ -0,0 +1,47 @@
...
+            email = fields[1][:-1]
+            identity['domain'] = email.split("@")[1]
+        except IndexError:
+            identity['domain'] = 'unknown'
+
+        return identity
+
+    @classmethod
+    def anonymize_item(cls, item):
+        """ Remove or hash the fields that contain personal information """
```
valeriocos: Please add a blank line between the docstring and the code (and remove the empty chars at the beginning and end of it). As general rule in GrimoireLab, one-line docstrings are separated with an empty line from the code, longer docstrings are not.

valeriocos: This applies also to the rest of the files in this PR :)

grimoire_elk/identities/git.py
valeriocos: This applies also to the rest of the files in this PR :)

valeriocos (commented):
Thank you @jjmerchante for providing a different implementation for #822 . Overall it looks great, there are some minor comments to address.

In this iteration we can skip the specific tests for the new files added (since they are covered by the tests you have provided).

I had a look at the setup.py and it looks it should be updated, I understand that we should add the new module.
```
packages=[..., 'grimoire_elk.identities`]
```

ref: https://github.com/chaoss/grimoirelab-elk/blob/master/setup.py#L59

valeriocos (approved):
LGTM, thanks @jjmerchante

PR #828, opened 2020-04-02
added projects_map db to improve coverage

imnitishng:
Noticed that many of the tests did not work completely because of missing projects json mapping. Added the line to use projects map from SQL database `test_projects.sql` ,observed increase in test coverage.


imnitishng:
Also, I see that `db_projects_map` has been deprecated. 
I saw no way to add another file similar to `projects.json` in our local machine for tests, so I went with this idea.

valeriocos (approved):
LGTM, thank you @imnitishng ! We are getting close to 80% !!

PR #830, opened 2020-04-03
[enriched] Check first if the instance of `item` is dict

zhquan:
This code avoid the error
`TypeError: argument of type 'NoneType' is not iterable`.


valeriocos (approved):
LGTM, thanks @zhquan

PR #728, opened 2019-11-09
[dependencies] Update redis to 3.0.0

valeriocos:
This code updates the dependecy for redis package, which is now pinned to 3.0.0

Related to https://github.com/chaoss/grimoirelab-elk/issues/726

PR #831, opened 2020-04-03
[gitter] add gitter backend support to ELK

### LINKED ISSUE #820: [gitter] Adding data storage for gitter
Since a new backend `gitter`  https://github.com/chaoss/grimoirelab-perceval/pull/632 has been added to perceval backend I was wondering if we can store the raw data in Elasticsearch and proceed forward to the implementation of this backend in `sir-mordred`?
I can start working on this. Please inform me if there some complications I'm missing.
WDYT @valeriocos?

imnitishng:
This PR is for adding an early support of gitter backend in grimoirelab-elk. Please have a look.
Fixes https://github.com/chaoss/grimoirelab-elk/issues/820


imnitishng:
I have added Raw and Enricher classes for the backend, however the enrichment of data is quite shallow at the moment because of the limited data returned by perceval. 
Studies are not implemented in this commit. I think it will be better to add them after some discussions, so please drop some ideas on that as well as creation of enriched data. 
I will be adding tests and documentation soon.

valeriocos:
Hi @imnitishng, sorry for the late reaction. I'm reviewing the PR right now

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,144 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2015-2019 Bitergia
```
valeriocos, line 3: Please update the last date

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,144 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#   Alvaro del Castillo San Felix <acs@bitergia.com>
```
valeriocos, line 19: Remove this author and add your info :)

grimoire_elk/raw/gitter.py
```diff
@@ -0,0 +1,84 @@
...
+
+class GitterOcean(ElasticOcean):
+    """Mattermost Ocean feeder"""
+
+    mapping = Mapping
+
+    @classmethod
+    def get_perceval_params_from_url(cls, url):
+        """ Get the perceval params given a URL for the data source """
+        params = []
```
valeriocos, line 78: Add an empty line between the docstring and the first statement

grimoire_elk/raw/gitter.py
```diff
@@ -0,0 +1,84 @@
...
+                        }
+                    }
+            }
+            '''
+
+        return {"items": mapping}
+
+
+class GitterOcean(ElasticOcean):
+    """Mattermost Ocean feeder"""
```
valeriocos, line 71: Please update the name

grimoire_elk/raw/gitter.py
```diff
@@ -0,0 +1,84 @@
...
+
+        Non dynamic discovery of type for:
+            * data.attachments.ts
+
+        :param es_major: major version of Elasticsearch, as string
+        :returns:        dictionary with a key, 'items', with the mapping
+        """
+
+        mapping = '''
+             {
```
valeriocos, line 40: Please change the mapping to:
```
        mapping = '''
             {
                "dynamic":true,
                "properties": {
                    "data": {
                        "dynamic":false,
                        "properties": {}
                    }
                }
            }
            '''
```

In this way, the index will take less storage. If you want to read more about this new mapping, please check:
- https://github.com/chaoss/grimoirelab-perceval/pull/554#issuecomment-513442373
- https://github.com/chaoss/grimoirelab-elk/pull/665

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,144 @@
...
+                }
+           }
+        } """
+
+        return {"items": mapping}
+
+
+class GitterEnrich(Enrich):
+
+    # This enricher must compatible with the Slack enricher to reuse the Kibiter panel
```
valeriocos, line 60: Why this comment? It is probably convenient to create a new panel for Gitter data. WDYT?

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,144 @@
...
+
+        for f in self.RAW_FIELDS_COPY:
+            if f in item:
+                eitem[f] = item[f]
+            else:
+                eitem[f] = None
+
+        message = item['data']
+
+        eitem["reply_count"] = 0  # be sure it is always included
```
valeriocos, line 109: This can be removed

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,144 @@
...
+
+        copy_fields = ["text", "unread", "readBy", "fromUser", "issues", "mentions",
```
valeriocos, line 111: `fromUser` isn't needed since this info is added when including the identity info. The attribute `unread` could be removed from the list, since it must be processed before.

What is stored in `issues`, `mentions`, `meta` and `urls`?

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,144 @@
...
+                       "meta", "urls", "id"]
+        for f in copy_fields:
+            if f in message:
+                eitem[f] = message[f]
+            else:
+                eitem[f] = None
+
+        eitem['text_analyzed'] = eitem['text']
```
valeriocos, line 119: I would add also the `sent` attribute, which can be useful for some visualizations. WDYT?

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,144 @@
...
+
+        eitem = self.__convert_booleans(eitem)
```
valeriocos, line 121: AFAIS there is just one boolean attribute (`unread`), so I'm not sure about including the method `convert_booleans`. We could have something like:
```
eitem['unread'] = 1 if message['unread'] else 0
```

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,144 @@
...
+        from_ = item
+
+        if identity_field == None:
+            identity['username'] = from_['data']['fromUser']['username']
+            identity['name'] = from_['data']['fromUser']['displayName']
+        else:
+            identity['username'] = from_['data'][identity_field]['username']
+            identity['name'] = from_['data'][identity_field]['displayName']
+
+        identity['email'] = None
```
valeriocos, line 84: this doesn't seem to be needed, since the email identity is initialized by default with `None`

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def get_sh_identity(self, item, identity_field=None):
+        # email not available for gitter
+        identity = {
+            'username': None,
+            'name': None,
+            'email': None
+        }
+
+        from_ = item
```
valeriocos, line 75: I understand the logic below, but it seems a bit reduntant since there is only an identity_field (i.e., `fromUser`).

What you think about the one below?
```
        if self.get_field_author() not in item['data']:
            return identity
        from_ = item['data'][self.get_field_author()]

        identity['username'] = from_.get('username', None)
        identity['name'] = from_.get('displayName', None)
        
        return identity
```

grimoire_elk/raw/gitter.py
```diff
@@ -0,0 +1,84 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#   Alvaro del Castillo San Felix <acs@bitergia.com>
```
valeriocos, line 19: Remove this name and add your info :)

grimoire_elk/raw/gitter.py
```diff
@@ -0,0 +1,84 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2015-2019 Bitergia
```
valeriocos, line 3: Update the copyright, thanks!

grimoire_elk/raw/gitter.py
```diff
@@ -0,0 +1,84 @@
...
+from ..elastic_mapping import Mapping as BaseMapping
+
+
+class Mapping(BaseMapping):
+
+    @staticmethod
+    def get_elastic_mappings(es_major):
+        """Get Elasticsearch mapping.
+
+        Non dynamic discovery of type for:
```
valeriocos, line 32: Please update the docstring

valeriocos (commented):
Thank you @imnitishng for the PR. Overall it looks great, I left a first round of comments :)

Please remember to
- include a schema of the enriched data (as a reference you can check the one for [git](https://github.com/chaoss/grimoirelab-elk/blob/master/schema/git.csv))
- update the mordred documentation to include this new data source (https://github.com/chaoss/grimoirelab-sirmordred#supported-data-sources-)

grimoire_elk/enriched/gitter.py
imnitishng: `mentions`stores the info about users mentioned (one list item per user mentioned)
```
 <class 'list'>: [{'screenName': 'IbrahimPatel89', 'userId': '5e870448d73408ce4fdf18b6', 'userIds': []}]
```

`issues` stores the info about issue mentioned in the message as a list of dictionary (number corresponds to issue number)
```
<class 'list'>: [{'repo': 'jenkinsci/jenkins', 'number': '3861'}, {'repo': 'jenkinsci/jenkins', 'number': '3863'}]
```

`url` has the URLs mentioned in the message
```
<class 'list'>: [{'url': 'https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fjira-plugin/detail/remove-deprecations/4/pipeline/56'}]
```

I have not yet encountered any message that populated the `meta` field, will update you once I know what it returns, will try searching the API docs.

grimoire_elk/enriched/gitter.py
imnitishng: Done! Thanks! :smile:

grimoire_elk/raw/gitter.py
imnitishng: Thank You, I missed that. Updated the mappings now.

grimoire_elk/enriched/gitter.py
imnitishng: Okay. Sounds good. I will remove it and have a discussion about new panels.

grimoire_elk/enriched/gitter.py
imnitishng: I'm sorry, I did not get this, what will the `sent` attribute store? The number of messages sent by  the particular user?

grimoire_elk/enriched/gitter.py
imnitishng: Seems good, thanks.

grimoire_elk/enriched/gitter.py
imnitishng: I would like to add, I found that the data returned from gitter API does not have email in it. So email will always be `None`, is it okay?

grimoire_elk/enriched/gitter.py
imnitishng: Looks good, thanks! Updated.

grimoire_elk/raw/gitter.py
imnitishng: Done! Thanks! :smile:

grimoire_elk/raw/gitter.py
imnitishng: Done!

grimoire_elk/raw/gitter.py
imnitishng: Done!

imnitishng:
I have resolved the issues and created schema for gitter. Please have a look. I just realised some of the attributes need to be removed from the schema I developed. I will do it soon and add the required tests for this backend. Thank You for the review!

imnitishng:
@valeriocos, I also wanted to discuss about the kinds of studies and visualizations we could perform in the data.
So far what I have in mind based on the panels for Git dashboard were -

- Count of authors in a room
- Count of issues and the issue submitters
- People being mentioned the most
- Messages by time zones
- Developers being inactive
- Attracted developers
- Active users
- Median time for replies
- Average people seeing the messages plotted over time
- Latest or Oldest URLs or issues being mentioned

This is a very high level idea of the future support I have planned, I will be digging into the details to implement this very soon. Please provide some suggestions or references that you feel I must look through. Thank you!

grimoire_elk/enriched/gitter.py
valeriocos: Sorry for not being clear. From the tests I did, `sent` (which is included in the data returned by the API) stores the date when the message was sent

imnitishng: Oh okay, I will include it. Thanks.

grimoire_elk/enriched/gitter.py
valeriocos: Thank you for the details @imnitishng !

I don't fully understand the data stored in each `mention`. `ScreenName` is the user mentioned or mentioning others? What is the difference between `userId` and `userIds`?

Beyond these questions, which data you think should be included to the enriched index? Do you have some visualizations in mind that can be implemented with these attributes?

grimoire_elk/enriched/gitter.py
valeriocos: Yes, that's perfect!

grimoire_elk/enriched/gitter.py
imnitishng: Hi! Sorry for unclear explanation.
So `mentions` is basically a list, with each list index corresponding to a dictionary -
The keys of the dictionary are -
`sceenName` - Displayed username of the user 
`userId` - User ID of the person mentioned
`userIds` - Apparently this is a key with an empty list as its value. For all the examples I've tested on, this attribute always remained empty. So it is useless as of now because it is just there and never fills up.

valeriocos:
Thank you @imnitishng for sharing an initial set of metrics! Can you answer the following questions?

> Count of issues and the issue submitters

How can we get the issue submitters?

> People being mentioned the most

It's a really intersting objective. If we go to this direction, I understand that we should expand the get_identities method to collect the user info stored in `mentions`. Does it make sense to measure also `people mentioning the most`?

> Developers being inactive, Attracted developers

How would you like to implement these metrics?

grimoire_elk/enriched/gitter.py
valeriocos: Thank you for your quick answer @imnitishng !
Is the `screenName` the one of the user mentioned?
Is there any difference between `displayName` and `screenName`?

imnitishng:
> How can we get the issue submitters?

Well issue submitters are the people mentioning some issue in the messages, so we already have the `userid` and `username` of the individual who has mentioned issues in his message. We could easily develop some rich data associated with the particular user ID. Just like how Git backend implements the number of issues or pull requests opened by a user.  

> people mentioning the most

Yea that sounds great, we could add it too, it might be a good metric to convey the most active users in a room

>Developers being inactive, Attracted developers
 How would you like to implement these metrics?

Well I had this idea as I went through git and github dashboards, I suppose people coming and joining the rooms over some defined time period (like GSoC time period of 1-2 months) might be classified in the category of attracted developers. 
The people going inactive without any recent messages or mentions in the chat room (again in a specified time period) might be considered as inactive.
This is a very non technical overview, however thanks for the ideas I will dive deeper into the codebase for implementing these metrics and get back to you with an initial implementation.

grimoire_elk/enriched/gitter.py
imnitishng: Yes, `screenName` is the username of the user mentioned. 
Yes, `displayName` is the real name of the user, whereas `sceenName` is the username. 

For ex, 
A user named **James Howe** has his gitter username as **OrangeDog** and the ID (userID) associated with this username something like this hash **647352fsd7723**
the attributes for this entry will be -

When this user send a message in the room- 
`displayName` - James Howe
`username` - OrangeDog
`url` - \OrangeDog
`id` - 647352fsd7723

When this user is mentioned in the room -
`screenName` - OrangeDog
`userId` - 647352fsd7723

valeriocos: Great, thank you for the details @imnitishng !

What do you think about processing the list of `mentions` to keep only the `userId` and `screenName`?

Since the idea is to have metrics that leverage on mention-related data, it could be convenient to store the `displayName` and `userId` of the author of the message in the enriched index as well. WDYT?

valeriocos:
Thank you for the clarifications @imnitishng !

> Well issue submitters are the people mentioning some issue in the messages

Is an approximation you are doing or does Gitter convert the issue submitters (the ones that opened an issue in GitHub) to users that mention those issues in their Gitter messages?

In the first case, we can redefine the metrics as `Count of issues and the users mentioning issues`

To move things forward, I would propose to work on a first implementation of the enricher/dashboard that focuses on a limited set of metrics among the ones you proposed. This set shouldn't require the implementation of a study. Later, we can see how to incorporate more metrics. WDYT?

grimoire_elk/enriched/gitter.py
imnitishng: > What do you think about processing the list of mentions to keep only the userId and screenName?

Yes sure, I will remove the unwanted data.

> it could be convenient to store the displayName and userId of the author of the message in the enriched index as well.

Okay, I will start working on creating the initial set of enriched data based on these suggestions and get you.

imnitishng:
> Is an approximation you are doing or does Gitter convert the issue submitters (the ones that opened an issue in GitHub) to users that mention those issues in their Gitter messages?

Gitter does not do anything as such, it just returns the data consisting of -

- The github repository whose issue has been mentioned
- The issue number (eg. #728)
- The name and id of the person who has mentioned the issue in his message.

There is no way of knowing who opened the issue in the github repository from the data returned by gitter API. However we could rely on some new functions based on github perceval backend to get info about the person who opened the issue.

> In the first case, we can redefine the metrics as Count of issues and the users mentioning issues

Yes this seems better. I'll do it.

> I would propose to work on a first implementation of the enricher/dashboard that focuses on a limited set of metrics

Sure, I am on it, will update the PR once I have the basic implementation complete. Thanks for the valuable suggestions.

valeriocos:
Thank you @imnitishng , ping me when the PR is ready for review

imnitishng:
Hi @valeriocos, I have added the metrics and enriched data. Please have a look and provide your suggestions.
I have the dashboard ready too, will make a PR for that in sigils too. Please have a look.

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,202 @@
...
+        for f in self.RAW_FIELDS_COPY:
+            if f in item:
+                eitem[f] = item[f]
+            else:
+                eitem[f] = None
+
+        message = item['data']
+
+        eitem['unread'] = 1 if message['unread'] else 0
+        eitem['text_html'] = message['html']
```
valeriocos, line 123: This isn't needed, since we have already the `text` attribute

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,202 @@
...
+        if self.prjs_map:
+            eitem.update(self.get_item_project(eitem))
+
+        eitem.update(self.get_grimoire_fields(item["metadata__updated_on"], "message"))
+
+        self.add_repository_labels(eitem)
+        self.add_metadata_filter_raw(eitem)
+        return eitem
+
+    def __get_rich_links(self, item):
```
valeriocos, line 154: why this method is private and the ones inside are not?

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,202 @@
...
+            else:
+                eitem[f] = None
+
+        message = item['data']
+
+        eitem['unread'] = 1 if message['unread'] else 0
+        eitem['text_html'] = message['html']
+
+        copy_fields = ["text", "readBy", "issues", "mentions",
+                       "meta", "urls", "id"]
```
valeriocos, line 126: "issues", "mentions", "meta", "urls" may not be copied, since "meta" is always empty and the other attributes are processed afterwards. Am I missing something?

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,202 @@
...
+            rich_item.update(self.enrich_mentions(item['mentions']))
+
+        if item['urls']:
+            url_parsed = urlparse(item['urls'][0]['url'])
+            rich_item['url_hostname'] = '{uri.scheme}://{uri.netloc}/'.format(uri=url_parsed)
+
+        return rich_item
+
+    def enrich_issues(self, issue_pr, html_text):
+        """Enrich issues or PRs mentioned in the message"""
```
valeriocos, line 171: I understand that this method is extracting issues/pull requests, maybe it is convenient to call it `extract_issues`, WDYT?

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,202 @@
...
+
+        rich_issue_pr = {}
+        links_found = self.HTML_LINK_REGEX.findall(html_text)
+
+        if 'repo' in issue_pr[0].keys():
```
valeriocos, line 176: Can you explain the logic of this method? Why only the first issue_pr is used?

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,202 @@
...
+                rich_issue_pr['github_url'] = links_found[0]
+                rich_issue_pr['repo'] = issue_pr[0]['repo']
+
+        return rich_issue_pr
+
+    def enrich_mentions(self, mentioned):
+        """Enrich users mentioned in the message"""
+
+        rich_mentions = {}
+        usrname = mentioned[0]['screenName']
```
valeriocos, line 191: Can you explain the logic of this method? Why only the first `mentioned` is used?

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,202 @@
...
+        copy_fields = ["text", "readBy", "issues", "mentions",
+                       "meta", "urls", "id"]
+
+        for f in copy_fields:
+            if f in message:
+                eitem[f] = message[f]
+            else:
+                eitem[f] = None
+
+        if message['fromUser']['username'] not in self.username_mapping:
```
valeriocos, line 134: I understand the usefulness of the `username_mapping`, however its advantage occurs only during the intial enrichment where thousands of messages are processed. For other enrichments, the dictionary may not be useful. It could be convenient to store only the username/screename (the information that directly available). Additional information could be added with a study (if needed). WDYT?

schema/gitter.csv
```diff
@@ -0,0 +1,42 @@
...
+offset
+origin,keyword,true,"Original URL where the room was retrieved from."
+project_1,keyword,true,"Project (if more than one level is allowed in project hierarchy)"
+project,keyword,true,"Project."
+readBy,int,true,"Count of users who have read the message"
+repository_labels,keyword,true,"Custom repository labels defined by the user."
+tag,keyword,true,"Perceval tag."
+text,keyword,true,"Message body in plain text. "
+text_html,keyword,true,"Message body in HTML format. "
+unread,boolean,true,"True if the message is unread by user, False if read. "
```
valeriocos, line 40: Who is the `user`? If it's the Perceval backend, I understand that this value should be always 0 or 1, so it isn't useful to be stored. Am I missing something?

schema/gitter.csv
```diff
@@ -0,0 +1,42 @@
...
+metadata__updated_on,date,true,"Date when the item was updated in its original data source."
+offset
+origin,keyword,true,"Original URL where the room was retrieved from."
+project_1,keyword,true,"Project (if more than one level is allowed in project hierarchy)"
+project,keyword,true,"Project."
+readBy,int,true,"Count of users who have read the message"
+repository_labels,keyword,true,"Custom repository labels defined by the user."
+tag,keyword,true,"Perceval tag."
+text,keyword,true,"Message body in plain text. "
+text_html,keyword,true,"Message body in HTML format. "
```
valeriocos, line 39: This could be removed since the same information is included in `text`

schema/gitter.csv
```diff
@@ -0,0 +1,42 @@
...
+unread,boolean,true,"True if the message is unread by user, False if read. "
+urls,list,true,"List of URLs mentioned in the message, Empty if no URLs found. "
```
valeriocos, line 41: The shape of URLs is the one below. It would be convenient to process it to have a flat list of strings.
```
"urls" : [
            {
              "url" : "https://javadoc.jenkins-ci.org/hudson/model/Run.html#getTimestamp--"
            },
            {
              "url" : "https://javadoc.jenkins-ci.org/hudson/model/Run.html#getTime--"
            }
          ],
```

schema/gitter.csv
```diff
@@ -0,0 +1,42 @@
...
+fromUser_name,keyword,true,"Sender name."
+fromUser_org_name,keyword,true,"Sender organization name from SortingHat profile."
+fromUser_user_name,keyword,true,"Sender username for the platform"
+fromUser_uuid,keyword,true,"Sender UUID from SortingHat."
+grimoire_creation_date,date,true,"Message date (when the original author sent the message)."
+id,keyword,true,"Sender user ID for gitter. "
+is_gitter_message,boolean,true,"True if the item is a Gitter message. "
+issues,list,true,"List of issues mentioned in the message, Empty if no issues were mentioned. "
+mentions,list,true,"List of users mentioned in the message, Empty if no user is mentioned. "
+meta,list,true,"Stores metadata for message, usually empty"
```
valeriocos, line 25: If it always empty, I would remove it.

schema/gitter.csv
```diff
@@ -0,0 +1,42 @@
...
+fromUser_multi_org_names,keyword,true,"List of the sender organizations from SortingHat profile."
+fromUser_name,keyword,true,"Sender name."
+fromUser_org_name,keyword,true,"Sender organization name from SortingHat profile."
+fromUser_user_name,keyword,true,"Sender username for the platform"
+fromUser_uuid,keyword,true,"Sender UUID from SortingHat."
+grimoire_creation_date,date,true,"Message date (when the original author sent the message)."
+id,keyword,true,"Sender user ID for gitter. "
+is_gitter_message,boolean,true,"True if the item is a Gitter message. "
+issues,list,true,"List of issues mentioned in the message, Empty if no issues were mentioned. "
+mentions,list,true,"List of users mentioned in the message, Empty if no user is mentioned. "
```
valeriocos, line 24: Mentions are processed by `enrich_mentions`, so it seems that this info is duplicated.

schema/gitter.csv
```diff
@@ -0,0 +1,42 @@
...
+fromUser_id,keyword,true,"Sender Id from SortingHat."
+fromUser_multi_org_names,keyword,true,"List of the sender organizations from SortingHat profile."
+fromUser_name,keyword,true,"Sender name."
+fromUser_org_name,keyword,true,"Sender organization name from SortingHat profile."
+fromUser_user_name,keyword,true,"Sender username for the platform"
+fromUser_uuid,keyword,true,"Sender UUID from SortingHat."
+grimoire_creation_date,date,true,"Message date (when the original author sent the message)."
+id,keyword,true,"Sender user ID for gitter. "
+is_gitter_message,boolean,true,"True if the item is a Gitter message. "
+issues,list,true,"List of issues mentioned in the message, Empty if no issues were mentioned. "
```
valeriocos, line 23: Issues are processed by `enrich_issues`, so this attribute seems redundant.

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,202 @@
...
+                if links_found[0].split('/')[-2] == 'issues':
+                    rich_issue_pr['issue'] = issue_pr[0]['repo'] + ' #' + issue_pr[0]['number']
+                else:
+                    rich_issue_pr['pull'] = issue_pr[0]['repo'] + ' #' + issue_pr[0]['number']
+                rich_issue_pr['github_url'] = links_found[0]
+                rich_issue_pr['repo'] = issue_pr[0]['repo']
+
+        return rich_issue_pr
+
+    def enrich_mentions(self, mentioned):
```
valeriocos, line 187: The method seems to extract the mentions, thus the name should be extract_mentions (or something similar).

valeriocos (commented):
Thank you @imnitishng for working on this PR. I have some questions about the method `__get_rich_links` and the ones related to it. Beyond this the PR looks good.

Can you add some tests? You can look at the existing tests (e.g., https://github.com/chaoss/grimoirelab-elk/blob/master/tests/test_slack.py) to write the ones for Gitter. 

The tests leverage on a common base (ref [here](https://github.com/chaoss/grimoirelab-elk/blob/master/tests/base.py)). In the test data [folder](https://github.com/chaoss/grimoirelab-elk/tree/master/tests/data), you need to add a file containing some docs extracted from the Pagure backend of Perceval (see example [here](https://github.com/chaoss/grimoirelab-elk/blob/master/tests/data/functest.json)). These docs are then used for testing the code of the raw and enrich connectors.

Thanks!

grimoire_elk/enriched/gitter.py
imnitishng: I assume that you are talking about the `text_html` attribute.
It should be stored because the links we extract for issues, pull requests and URL are stored in HTML <a > anchor tags. Links are extracted using regex.


`text_html` for this `text` message is

> @slide This is what I'm hacking on: https://github.com/jenkinsci/configuration-as-code-plugin/issues/565


```
<span data-link-type="mention" data-screen-name="slide" class="mention">@slide</span> This is what I&#39;m hacking on: <a target="_blank" data-link-type="issue" data-issue="565" href="https://github.com/jenkinsci/configuration-as-code-plugin/issues/565" data-provider="github" data-issue-repo="jenkinsci/configuration-as-code-plugin" class="issue">jenkinsci/configuration-as-code-plugin#565</a>
```

I'm sorry, the implementation is also possible from `text` field, I just noticed it right now.
However, this attribute might be useful to extract info about the images or code formatting sent in messages.

grimoire_elk/enriched/gitter.py
imnitishng: Sorry, I did not wanted it to be private. I'll fix it.

grimoire_elk/enriched/gitter.py
imnitishng: Yes, they can be skipped but issues stores multiple issues mentioned in a single message so this might be kept. (explained below)
I will remove `meta` thanks.

grimoire_elk/enriched/gitter.py
imnitishng: Okay will do it.

grimoire_elk/enriched/gitter.py
imnitishng: Okay so, when people mention some issue or PR in message it is stored inside the `issues` attribute as a list of dictionary, each list index corresponds to a single issue mentioned.
Now initially I implemented the approach to collect an store multiple issues mentioned in a single message as a nested attribute in enriched data.
But while creating panels when I tried visualizing the most mentioned issue, by specifying attribute as `issues.issue_number` I ran into problems, because for a single message with 2 issues, the table created 4 entries (instead of 2).

Some thing like this
| No.  | Pull Request| count
|:----:|:--------------------------------------------------------------------------------------------------------------------------:|:---------:|
|jenkinsci/durable-task-plugin#81|https://github.com/jenkinsci/durable-task-plugin/pull/81|1
|jenkinsci/saml-plugin#78|https://github.com/jenkinsci/saml-plugin/pull/78|1|
|jenkinsci/durable-task-plugin#81|https://github.com/jenkinsci/saml-plugin/pull/78|1
|jenkinsci/saml-plugin#78|https://github.com/jenkinsci/durable-task-plugin/pull/81|1|

| No.  | Pull Request| count
|:----:|:--------------------------------------------------------------------------------------------------------------------------:|:---------:|
|jenkinsci/durable-task-plugin#81|https://github.com/jenkinsci/durable-task-plugin/pull/81|1
|jenkinsci/saml-plugin#78|https://github.com/jenkinsci/saml-plugin/pull/78|1|


Same happened with multiple mentions in a single message.
There was no way I could think of to store multiple issues or mentions in a single enriched item so that it created correct visualizations.
Hence I had to use only the first issue, PR, mention, URL in a single message.
I noticed people don't usually mention multiple users or issues in a single message, so I thought this would be the way to go. Would love to hear from you if you have a better way of implementing this in mind.

Here is the logic I used earlier to extract and store multiple issues in the enriched item.

```
        if eitem['fromUser_user_name'] not in username_mapping:
            username_mapping[eitem['fromUser_user_name']] = eitem['fromUser_name']

        if eitem['issues']:
            for issue in eitem['issues']:
                issue['real_name'] = eitem['fromUser_name']
                issue['full_issue'] = str(issue['repo'] + ' #' + issue['number'])

        if eitem['mentions']:
            for mention in eitem['mentions']:
                if mention['screenName'] in username_mapping:
                    mention['real_name'] = username_mapping[mention['screenName']]
                else:
                    mention['real_name'] = 'Unknown'
```

grimoire_elk/enriched/gitter.py
imnitishng: Same as above

grimoire_elk/enriched/gitter.py
imnitishng: Well this dictionary was added so that when viewing the dashboard with the most users mentioned, the columns show `username`, `real name` and `no. of times mentioned` since gitter API only returns `username` in the `mentions` attribute.
So for convenience of the user viewing the dashboards, I added the real name field so he/she does not have to lookup who belongs to this username.

Is it okay or do i remove this dictionary and the column from dashboard.

schema/gitter.csv
imnitishng: These schemas need a complete update. I will do it in the next iteration with tests. Thanks for pointing out.

schema/gitter.csv
imnitishng: Okay I will do it

schema/gitter.csv
imnitishng: Okay thanks

schema/gitter.csv
imnitishng: Sure :smile:

schema/gitter.csv
imnitishng: Thanks for pointing out

grimoire_elk/enriched/gitter.py
imnitishng: Sure

schema/gitter.csv
imnitishng: Will remove, thanks.

grimoire_elk/enriched/gitter.py
valeriocos: Yes I was referring to `text_html`, sorry for not being precise!

The attribute `text_html` can be used to derive info (e.g., pull requests and urls), but it shouldn't be stored in the enriched item if it isn't used by any visualization. The main reason of not storing it is to reduce the size of the enriched index.

> I'm sorry, the implementation is also possible from text field, I just noticed it right now.

No worries!

> However, this attribute might be useful to extract info about the images or code formatting sent in messages.

Yes, I understand, but the information can be taken from the raw data when it will be needed.  One of the advantages of having raw and enriched data is that the first one contains a lot of information, while the latter processes only a part (which is the one showed in the dashboards). Thus, in the specific case of images, the day we will decide to add image information, we will derive it from the raw data.

grimoire_elk/enriched/gitter.py
valeriocos: Thanks!

grimoire_elk/enriched/gitter.py
imnitishng: Okay, thanks I got your point. I will fix this. :smiley:

grimoire_elk/enriched/gitter.py
imnitishng: Updated the table columns, just noticed, they were not rendering properly

grimoire_elk/enriched/gitter.py
valeriocos: Thank you for the clarification! The risk with this approach is that the `real name` is correctly resolved only if that user has appeared in one of the documents processed during the enrichment phase. Thus, depending on the enrichment phase (and the documents processed in it), the real name can be set with the right info or with `Unknown`.

A solution is to initialize the dict `username_mapping` with the usernames and real names present in the enriched index, thus  `username_mapping` doesn't depend on the information of the single enrichment phase. Another solution is to implement a study, which is executed after the enrichment phase. However, to avoid adding too much complexity in this iteration, it may be convenient to show only the `username`. The real name resolution can be done afterwards.

WDYT?

grimoire_elk/enriched/gitter.py
valeriocos: Thanks!

grimoire_elk/enriched/gitter.py
imnitishng: Okay thanks, for now I will go with removing the dictionary and showing only the `usernames`, will think of mapping real name when we implement studies.

grimoire_elk/enriched/gitter.py
valeriocos: Thank you for the clarifications @imnitishng ! I modified the code as shown below. The attributes `issues`, `mentions` and `urls` are copied to the enriched item and then modified in-place.

WRT to the `enrich_issues`, it isn't always possible to identify issue urls, since the issue is referenced only with a hashtag (e.g., `#10`). 

WRT mentions, some times there are more attributes beyond the ones about the user such as `announcement` and `group`.

WRT urls, I don't know if it is useful to have them in the enriched indexes. How do you want to exploit them?

```
--- a/grimoire_elk/enriched/gitter.py
+++ b/grimoire_elk/enriched/gitter.py
@@ -122,8 +122,7 @@ class GitterEnrich(Enrich):
         eitem['unread'] = 1 if message['unread'] else 0
         eitem['text_html'] = message['html']
 
-        copy_fields = ["text", "readBy", "issues", "mentions",
-                       "meta", "urls", "id"]
+        copy_fields = ["text", "readBy", "issues", "mentions", "urls", "id"]
 
         for f in copy_fields:
             if f in message:
@@ -134,7 +133,17 @@ class GitterEnrich(Enrich):
         if message['fromUser']['username'] not in self.username_mapping:
             self.username_mapping[message['fromUser']['username']] = message['fromUser']['displayName']
 
-        eitem.update(self.__get_rich_links(item['data']))
+        if eitem['issues']:
+            self.enrich_issues(eitem['issues'], eitem['text'])
+
+        if eitem['mentions']:
+            self.enrich_mentions(eitem['mentions'])
+
+        if eitem['urls']:
+            for url in eitem['urls']:
+                url_parsed = urlparse(url['url'])
+                hostname = '{uri.scheme}://{uri.netloc}/'.format(uri=url_parsed)
+                url['hostname'] = hostname
 
         message_timestamp = str_to_datetime(eitem['metadata__updated_on'])
         eitem['tz'] = int(message_timestamp.strftime("%H"))
@@ -151,52 +160,32 @@ class GitterEnrich(Enrich):
         self.add_metadata_filter_raw(eitem)
         return eitem
 
-    def __get_rich_links(self, item):
-
-        rich_item = {}
-
-        if item['issues']:
-            rich_item.update(self.enrich_issues(item['issues'], item['html']))
-
-        if item['mentions']:
-            rich_item.update(self.enrich_mentions(item['mentions']))
-
-        if item['urls']:
-            url_parsed = urlparse(item['urls'][0]['url'])
-            rich_item['url_hostname'] = '{uri.scheme}://{uri.netloc}/'.format(uri=url_parsed)
-
-        return rich_item
-
-    def enrich_issues(self, issue_pr, html_text):
+    def enrich_issues(self, issues, text):
         """Enrich issues or PRs mentioned in the message"""
 
-        rich_issue_pr = {}
-        links_found = self.HTML_LINK_REGEX.findall(html_text)
-
-        if 'repo' in issue_pr[0].keys():
-            if links_found:
-                if links_found[0].split('/')[-2] == 'issues':
-                    rich_issue_pr['issue'] = issue_pr[0]['repo'] + ' #' + issue_pr[0]['number']
-                else:
-                    rich_issue_pr['pull'] = issue_pr[0]['repo'] + ' #' + issue_pr[0]['number']
-                rich_issue_pr['github_url'] = links_found[0]
-                rich_issue_pr['repo'] = issue_pr[0]['repo']
-
-        return rich_issue_pr
-
-    def enrich_mentions(self, mentioned):
+        for issue in issues:
+            issue['url'] = None
+            issue['type'] = None
+            try:
+                url_regexp = re.compile(
+                    r'.*(?P<platform_url>http.*)/{}/(?P<item_type>.*)/{}'.format(
+                        issue['repo'],
+                        issue['number']))
+            except:
+                continue
+
+            m = url_regexp.match(text)
+            if m:
+                platform = m.group('platform_url')
+                item_type = m.group('item_type')
+                issue['url'] = "{}/{}/{}/{}".format(platform, issue['repo'], item_type, issue['number'])
+                issue['type'] = 'issue' if item_type == 'issues' else 'pull request'
+
+    def enrich_mentions(self, mentions):
         """Enrich users mentioned in the message"""
 
-        rich_mentions = {}
-        usrname = mentioned[0]['screenName']
-
-        if usrname in self.username_mapping:
-            rich_mentions['mentioned_realname'] = self.username_mapping[usrname]
-        else:
-            rich_mentions['mentioned_realname'] = 'Unknown'
-        rich_mentions['mentioned_username'] = mentioned[0]['screenName']
-
-        if 'userId' in mentioned[0].keys():
-            rich_mentions['mentioned_userid'] = mentioned[0]['userId']
-
-        return rich_mentions
+        for mention in mentions:
+            mention['username'] = mention['screenName']
+            mention.pop('screenName', None)
+            mention.pop('userId', None)
+            mention.pop('userIds', None)
```

After generating the enriched index with the code above, I built the visualization below, which counts the number of times an issue/pull request appear in gitter message. I don't know if I'm missing something or this is the result you expect.

captura_326

imnitishng: Hi @valeriocos thank you for testing it out. 

> WRT to the enrich_issues, it isn't always possible to identify issue urls, since the issue is referenced only with a hashtag (e.g., #10).

Yes that is a problem, I have seen when the issue was referenced only by hashtag then the `repo` attribute inside `issues` is empty, so I made a check 
```python
if 'repo' in entity.keys():
    ...
```
since the issue URL cannot be extracted in this case.

> WRT mentions, some times there are more attributes beyond the ones about the user such as announcement and group.

Is it so? I never saw it in the raw data neither do i see it here
Screenshot from 2020-04-13 10-47-51

> WRT urls, I don't know if it is useful to have them in the enriched indexes. How do you want to exploit them?

For now I just added this in enriched index to produce a pi chart visualization in kibiter for the top mentioned URL hosts (eg. stackoverflow.com, github.com, etc)

> After generating the enriched index with the code above, I built the visualization below, which counts the number of times an issue/pull request appear in gitter message.

Thank you for taking the time out and testing this.
Yes this is the visualization I wanted but with one more column storing the URL of the respective issue or pull request.
I modified my code for supporting multiple mentioned issues in a single message, the visualization I got was this, 
Screenshot from 2020-04-13 10-33-29

And when I remove the bucket for URL, then I get this
Screenshot from 2020-04-13 10-33-21


Here is an example of the raw data for message with multiple PR mentions
Screenshot from 2020-04-13 10-59-10

valeriocos: You're welcome! 

What's the difference between the first and second visualization? They seem having the same information. What you would like to measure?

imnitishng: The second visualization has wrong URL values, if you see rows 1 and 2, the URL in second row is wrong, it should not be there.
Plus in rows 6-9, notice how PR URLs have been treated, (they are exchanged) since these PRs are mentioned in a single message.
What I want is 1 row for row 1 and 2 (row 2 should be removed)
And 2 rows instead of 4 starting from row 6.

imnitishng:
Hi @valeriocos, I have added the tests, please have a look. However some work in logic and schemas  is left because of the pending discussions. Will finalize the PR as soon as we are done with the discussion. Thank you for the help.

grimoire_elk/enriched/gitter.py
imnitishng: I gave this a thought and decided to not include the URL column and just go for the visualization you suggested [here](https://github.com/chaoss/grimoirelab-elk/pull/831#discussion_r407229952). Please let me know how I should go about this. Thanks.

valeriocos: Ok, perfect @imnitishng ! We can try to include that visualization in a second iteration.

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,192 @@
...
+
+        self.add_repository_labels(eitem)
+        self.add_metadata_filter_raw(eitem)
+        return eitem
+
+    def get_rich_links(self, item):
+
+        rich_item = {}
+
+        if item['issues']:
```
valeriocos, line 155: Why the methods called by `get_rich_links` consider only the first entries of their list?

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,192 @@
...
+    @staticmethod
+    def get_elastic_mappings(es_major):
+        """Get Elasticsearch mapping.
+
+        :param es_major: major version of Elasticsearch, as string
+        :returns:        dictionary with a key, 'items', with the mapping
+        """
+
+        mapping = """
+        {
```
valeriocos, line 50: This mapping doesn't seem to be used. You should copy the `text` in the attribute `text_analyzed` in the method get_rich_item (i.e., `eitem['text_analyzed'] = eitem['text']`)

valeriocos (commented):
Hi @imnitishng , overall the PR looks good. I left some minor comments. Thanks for your work

grimoire_elk/enriched/gitter.py
imnitishng: Yes I was aware of that, was waiting for your reply to confirm the approach to be followed, fixed it now.

imnitishng:
Hi @valeriocos, I have finished the work on this backend based on your suggestions and submitted PRs in the concerned repositories. Please have a look, thank you! :smiley:

grimoire_elk/enriched/gitter.py
```diff
@@ -0,0 +1,188 @@
...
+
+from .enrich import Enrich, metadata
+from ..elastic_mapping import Mapping as BaseMapping
+
+
+logger = logging.getLogger(__name__)
+
+# Global Variables
+# username_mapping = {}
+GITHUB = 'https://github.com/'
```
valeriocos, line 36: This doesn't seem to be used, can you remove it. Please remove also lines 34 and 35. Thanks!

valeriocos (commented):
Hi @imnitishng thank you the PR. I left a really minor comment.

If you are interested, you can add a note to this PR which will appear in the next release. The note should include a high-level description of the changes introduced in this PR (to let non technical people understand what the change is about).

If you are interested in adding the note, please follow the instructions at: https://github.com/Bitergia/release-tools#changelog. Consider also to share your feedback when using the tool (what you liked, what you didn't like and things you would like to improve).

If you're not interested, can you clarify why?

In both cases, your feedback is valuable and will allow us to improve the way to reflect the code changes in the release.

Thanks!

imnitishng:
@valeriocos done, please have a look.

valeriocos:
thanks @imnitishng ! Did you commit the release note file (https://github.com/chaoss/grimoirelab-elk/pull/831#pullrequestreview-393940712) ?

imnitishng:
Doing it now.

imnitishng:
Done!

valeriocos (approved):
LGTM, thanks @imnitishng for your patience and good work!

PR #847, opened 2020-04-14
[enriched-bugzillarest] Add new field `is_open`

zhquan:
This code add the field `is_open` to know if the tiket is open.


valeriocos (approved):
LGTM, thanks @zhquan

PR #848, opened 2020-04-15
[metadata] Remove metadata from empty enriched items

### LINKED ISSUE #846: [enrich] Update decorator to add metadeta only for non-empty items
The decorator present in the below method adds metadata to the enriched item even if the item is empty https://github.com/chaoss/grimoirelab-elk/blob/48486d09092f466b21572ac3b9182f4e6310c544/grimoire_elk/enriched/enrich.py#L84

An empty item may be generated in case the item does not belong to a category that is to be enriched. For e.g in case, the item does not belong to 'issues', 'pull requests' or 'repositories' in the case of github2 enricher.
https://github.com/chaoss/grimoirelab-elk/blob/48486d09092f466b21572ac3b9182f4e6310c544/grimoire_elk/enriched/github2.py#L219

An update is also needed here to upload only non-empty documents: 
https://github.com/chaoss/grimoirelab-elk/blob/48486d09092f466b21572ac3b9182f4e6310c544/grimoire_elk/enriched/github2.py#L389

animeshk08:
Fixes: #846 
This commit updates the enricher for
github and gitlab to prevent addition of
metadata fields to empty enriched items

Signed-off-by: Animesh Kumar <animuz111@gmail.com>


animeshk08:
Hi @valeriocos. I have checked the enrichers for all the backends. Even though most of the backend override the `enrich_items` method they do not require this fix as there no `category` based checks.
I other words in all the other backends except GitHub and GitLab the enrich items are never empty.

Please have a look at this PR :)

Also, I will update the tests and separate the commits tomorrow :)

animeshk08:
Hi, @valeriocos. Please have a look at this PR :)
SInce only GitHub and GitLab had such a category check of items the PR covers only these backends.
Do you think we should extend this check to other backends as well? This will serve two main purposes:
* In case a raw item is not to be enriched at all.
* In case fetching of a new category of item is integrated and work is being done on creating its enricher.
WDYT?

valeriocos:
Hi @animeshk08 , sorry for the late reply! ^^

We are testing it in the pre-production environment, and the initial tests are good.
 It will be probably go into the release 0.2.41 (not the upcoming one, but the next one).

animeshk08:
Thank you for the quick reply @valeriocos. Let me know when there are any updates.
We can extend the discussion on https://github.com/chaoss/grimoirelab-elk/pull/848#issuecomment-618975697 then :)

PR #849, opened 2020-04-16
[enriched-git] Handle connection problems during enrich_git_branches

valeriocos:
This code allows to handle connection problems when accessing the repo to update the git branches.


zhquan (approved):
LGTM

PR #851, opened 2020-04-17
[launchpad] Add support for launchpad

imnitishng:
This commit adds support for Launchpad perceval backend in ELK.
Raw and enriched classes have been added.


imnitishng:
Hi @valeriocos, here is the first iteration of launchpad backend for ELK.
Please have a look, will be adding schemas and tests soon.
Would like to hear from you what sort of enrichment or studies we can plan for this backend.
Also I observed, getting data from launchpad was a quite slow compared to other backend, so getting data from a huge source like [Ubuntu](https://bugs.launchpad.net/ubuntu), might be quite sluggish.

valeriocos:
Hi @imnitishng ! The start looks promising!

> Would like to hear from you what sort of enrichment or studies we can plan for this backend.

In this PR, I would suggest to limit the scope of the enriched data to be produced. The studies and further enriched attributes can be added later on. It would be also convenient to list some metrics to be visualized (as done here: https://github.com/chaoss/grimoirelab-elk/pull/831#issuecomment-609479044)

You can take a look at the following enrichers to see which is the typical information produced.
- https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/bugzillarest.py
- https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/gitlab.py
- https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/redmine.py

> Also I observed, getting data from launchpad was a quite slow compared to other backend, so getting data from a huge source like Ubuntu, might be quite sluggish.

Thank you for the info. To speed up the testing of this backend, we can focus on small repos. Once we have a mature implementation, we can test it on something like Ubuntu. WDYT?

imnitishng:
Sure, I'll get back to you with the metrics.

imnitishng:
Hi @valeriocos 
I have updated the PR, reduced the data stored in enriched indexes, added some new attributes. Please have a look and suggest any improvements.

Here are some metrics I have in mind -

For a general overview
- Count of bugs over time
- Bugs by status over time (triaged, new, confirmed, fix released)
- Submitters per month over time
- Most bugs submitted by users (table)
- Bugs per Organizations 
- Bugs by organizations over time

For timing study or in a different menu (like GitHub issues timing)
- Status widget with number of bugs, submitters, assignees, median time open
- Median time open, 80% time open
- Bugs and submitters over time
- Bugs by severity 
- Bugs count by their status 
- Time taken to fix (fix commit) a bug or time worked on (time duration from when the bug was marked in progress to fix committed)

A geolocation study can be an option giving us the visualization for

-  Assignee or owner by time zone

valeriocos:
Thank you @imnitishng for advancing on this pull request. Could you share some launchpad repos to have a common base to test the backend?

> https://github.com/chaoss/grimoirelab-elk/pull/851#issuecomment-616077984

The metrics look good! I would skip the last metric (`Assignee or owner by time zone`) using the geolocation study, to avoid adding too much complexity to this PR. We can include studies in a different pull request.

Thanks!

imnitishng:
Yes sure.
For now I have been testing on the [launchpad](https://launchpad.net/launchpad) repo itself.
We can have the [Terminator](https://bugs.launchpad.net/terminator), [MySQL](https://bugs.launchpad.net/mysql) repos as an option too.

>  I would skip the last metric (Assignee or owner by time zone) using the geolocation study, to avoid adding too much complexity to this PR.

Sure. Waiting on your suggestions for this iteration.

imnitishng:
Hi @valeriocos, I will start the work on panels and tests, would love to have a review so we can think upon adding or modifying some attributes.

valeriocos:
Thank you @imnitishng ! As soon as the work on the Gitter dashboard is completed, I'll review this PR. The idea behind this decision is to concentrate the effort in one thing, finalize it and move then to another task.

imnitishng:
Oh, okay sure. Thanks.

valeriocos:
Sorry @imnitishng , I'm late with this PR. I'll have a look at it at most tomorrow

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,225 @@
...
+            if f in item:
+                eitem[f] = item[f]
+            else:
+                eitem[f] = None
+
+        # data fields to copy
+        copy_fields = ["title", "web_link", "date_created", "date_incomplete", "is_complete",
+                       "status", "bug_target_name", "importance", "date_triaged", "date_left_new", "project"]
+
+        self.__copy_fields(copy_fields, item['data'], eitem)
```
valeriocos, line 112: It may be convenient to open a new PR to:
- move this method to the parent class https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/enrich.py#L104
- rename it to `copy_raw_fields`
- and refactor the code of the other enrichers where possible (https://github.com/chaoss/grimoirelab-elk/search?q=RAW_FIELDS_COPY&unscoped_q=RAW_FIELDS_COPY)

WDYT?

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,225 @@
...
+
+        rich_bugtask['time_worked_on'] = get_time_diff_days(data['date_in_progress'], data['date_fix_committed'])
+
+        rich_bugtask['time_to_confirm'] = get_time_diff_days(data['date_created'], data['date_confirmed'])
+
+        # Author and assignee data
+        owner = data.get('owner_data', None)
+        if owner is not None and owner:
+            rich_bugtask['user_login'] = owner['name']
+            rich_bugtask['author_name'] = owner['display_name']
```
valeriocos, line 169: why `author_name`? (same comments applies for other similar cases below)

valeriocos (commented):
Hi @imnitishng sorry for the late review. Overall the code looks good and easy to ready, I left some comments.

I have tested the PR with:
```
setup.cfg
[launchpad]
raw_index = launchpad_raw_test
enriched_index = launchpad_enrich_test

projects.json
"launchpad": [
            "https://launchpad.net/launchpad"
        ]
```

Please check also the comments below:

- Note that the Perceval backend (ref [here](https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/launchpad.py#L70)) allows also to fetch data from a package in a distribution. This case doesn't seem to be covered by the current implementation

- Rebase your PR, it isn't aligned with `master`

- Start adding tests when you have time, the review process will be easier.

- Update the PR description with the repos you are using (or plan to) to test the PR.

- Update the documentation in ELK and mordred

Thanks!

imnitishng:
@valeriocos Should I open a new PR, cannot get rid of this problem after I did a rebase. Please help.

valeriocos:
Hi @imnitishng , yes open a new PR, if you cannot advance. However, it is convenient to use this PR to get familiar with the rebase.

Can you list the steps you did when rebasing this PR? Thanks!

imnitishng:
Thank you @valeriocos 
Here is what I did, 

```
git fetch upstream
git rebase upstream/master
```

Realised I had not committed after adding tests
```
git commit -m "added tests and schemas" --signoff 
git rebase upstream/master
git rebase -i master
```
Squashed commits
```
git push --force-with-lease origin launchpad_elk 
```
I think it happened because I fetched before committing?

imnitishng:
I get the problem, trying to fix it.

valeriocos:
I guess that the problem could have been with the last instruction: `git rebase -i master`. 
Once you execute `git rebase upstream/master`, an editor should be prompted to fix the conflicts. Once done, you don't need to rebase with master.

grimoire_elk/enriched/launchpad.py
imnitishng: Sure. I will send a PR as soon as this backend is added.

grimoire_elk/enriched/launchpad.py
imnitishng: Removed

imnitishng:
Well okay. I also fixed this PR.
@valeriocos, sorry for all the hassle :sweat_smile: but what do we do now? 
Continue with this or the new PR? 
Tests here will fail, will update the PR you choose to work on.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,225 @@
...
+
+        identity = self.get_sh_identity(item)
+        yield identity
+
+    @metadata
+    def get_rich_item(self, item):
+
+        eitem = {}
+
+        for f in self.RAW_FIELDS_COPY:
```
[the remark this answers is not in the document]
imnitishng: Removed :smile:

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,225 @@
...
+        from_ = item
+        if isinstance(item, dict) and 'data' in item:
+            if item['data'][self.get_field_author()]:
+                from_ = item['data'][self.get_field_author()]
+                identity['username'] = from_['name']
+                identity['name'] = from_['display_name']
+
+        return identity
+
+    def get_identities(self, item):
```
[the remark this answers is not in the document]
imnitishng: Sorry, I have updated it, did not notice this the last time.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,225 @@
...
+    def get_elastic_mappings(es_major):
+        """Get Elasticsearch mapping.
+
+        geopoints type is not created in dynamic mapping
+
+        :param es_major: major version of Elasticsearch, as string
+        :returns:        dictionary with a key, 'items', with the mapping
+        """
+
+        mapping = '''
```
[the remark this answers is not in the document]
imnitishng: Updated.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,225 @@
...
+
+        for f in self.RAW_FIELDS_COPY:
+            if f in item:
+                eitem[f] = item[f]
+            else:
+                eitem[f] = None
+
+        # data fields to copy
+        copy_fields = ["title", "web_link", "date_created", "date_incomplete", "is_complete",
+                       "status", "bug_target_name", "importance", "date_triaged", "date_left_new", "project"]
```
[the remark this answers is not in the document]
imnitishng: That was for testing purpose. Removed it now.

valeriocos:
Hi @imnitishng , no worries ! I'm fine with any of the two PRs. Please close one and I'll review the other one.

Thanks!

imnitishng:
Let's go with this one. Thanks.

valeriocos:
Hi @imnitishng , can you rebase your PR to include the changes of #864? Thanks!

imnitishng:
Done :smile:

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+        rich_bugtask['time_to_close_days'] = \
+            get_time_diff_days(data['date_created'], data['date_closed'])
+
+        if not data['date_closed'] and not data['is_complete']:
+            rich_bugtask['time_still_open_days'] = \
+                get_time_diff_days(data['date_created'], datetime_utcnow().replace(tzinfo=None))
+
+        if not data['date_closed']:
+            rich_bugtask['time_open_days'] = \
+                get_time_diff_days(data['date_created'], datetime_utcnow().replace(tzinfo=None))
```
valeriocos, line 143: `time_open_days` is not correctly calculated, since this field should be updated every day. The way of achieving this is by using a scripted field in the dashboard.

Imagine that we are fetching issues from the ubuntu issue tracker. In the first execution, `time_open_days` will be correctly calculated, since it will be derived by comparing the issue creation date and the day when the execution was made. However this field will be recalculated only when the issue is updated upstream. Thus, the `time_open_days` won't reflect the metric `time_open_days`, but just the time when the issue was last updated.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+            if not data['is_complete']:
+                rich_bugtask['time_still_open_days'] = rich_bugtask['time_open_days']
+        else:
+            rich_bugtask['time_open_days'] = rich_bugtask['time_to_close_days']
```
valeriocos, line 147: Can you clarify it this statement?

I understand that:
- `time_to_close_days` is the time to close an issue (from its creation to the closing action). It is applied to issues that have been closed.
- `time_open_days` is the time an issue is open (from its creation until today). It is applied to issues that are not closed.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+        return rich_bugtask
+
+    def __extract_bug_info(self, bug_data):
+
+        rich_bug_info = {}
+
+        copy_fields = ["latest_patch_uploaded", "security_related", "private", "users_affected_count",
+                       "title", "description", "tags", "date_last_updated", "message_count", "heat"]
+
+        rich_bug_info['time_to_last_update_days'] = \
```
valeriocos, line 211: time_to_last_update_days is also calculated at line 150. This statement is more coherent with the name of the metric.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+        if not data['date_closed']:
+            rich_bugtask['time_open_days'] = \
+                get_time_diff_days(data['date_created'], datetime_utcnow().replace(tzinfo=None))
+            if not data['is_complete']:
+                rich_bugtask['time_still_open_days'] = rich_bugtask['time_open_days']
+        else:
+            rich_bugtask['time_open_days'] = rich_bugtask['time_to_close_days']
+
+        if data['activity_data']:
+            rich_bugtask['time_to_last_update_days'] = \
```
valeriocos, line 150: Why `time_to_last_update_days` is calculated in two different ways (see comment at line 211) ?

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def __get_rich_bugs(self, data):
+        """Create enriched data for bugs"""
+
+        rich_bugtask = {}
+
+        # Time to
+        rich_bugtask["created_to_assigned"] = get_time_diff_days(data['date_created'], data['date_assigned'])
+        rich_bugtask['assigned_to_closed'] = get_time_diff_days(data['date_assigned'], data['date_closed'])
+        rich_bugtask["time_to_close_days"] = get_time_diff_days(data['date_created'], data['date_closed'])
```
valeriocos, line 132: This should be calculated only for closed issues, while for open issues we should have time_open_days

grimoire_elk/raw/launchpad.py
```diff
@@ -16,13 +16,47 @@
...
 class LaunchpadOcean(ElasticOcean):
     """Launchpad Ocean feeder"""
 
-    pass
+    mapping = Mapping
+
+    @classmethod
+    def get_perceval_params_from_url(cls, url):
+        params = []
+        distribution = url.split('/', 3)[-1]
```
valeriocos, line 59: We should take into account also repositories including the `package` (https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/launchpad.py#L70). Is there a specific reason to discard them?

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+
+        rich_bugtask['time_to_fix_commit'] = get_time_diff_days(data['date_created'], data['date_fix_committed'])
+
+        rich_bugtask['time_worked_on'] = get_time_diff_days(data['date_in_progress'], data['date_fix_committed'])
+
+        rich_bugtask['time_to_confirm'] = get_time_diff_days(data['date_created'], data['date_confirmed'])
+
+        # Author and assignee data
+        owner = data.get('owner_data', None)
+        if owner is not None and owner:
```
valeriocos, line 163: this could be just `if owner`

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+            rich_bugtask['user_login'] = owner['name']
```
valeriocos, line 164: please replace the access to the attributes with get. E.g, owner.get('name', None), this will prevent possible key errors.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+            rich_bugtask['user_name'] = owner['display_name']
+            rich_bugtask['user_joined'] = owner['date_created']
+            rich_bugtask['user_karma'] = owner['karma']
+            rich_bugtask['user_time_zone'] = owner['time_zone']
+        else:
```
valeriocos, line 169: The else condition can be removed as follows. This would simplify the test cases (currently line 170-174 are not tested).
```
rich_bugtask['user_name'] = None
...
owner = data.get('owner_data', None)
if owner:
   rich_bugtasks['user_name'] = owner.get('name', None)
   ...
```

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+            rich_bugtask['user_name'] = None
+            rich_bugtask['user_realname'] = None
+            rich_bugtask['user_joined'] = None
+            rich_bugtask['user_karma'] = None
+            rich_bugtask['user_time_zone'] = None
+
+        assignee = data.get('assignee_data', None)
```
valeriocos, line 176: Same comments as above

valeriocos (commented):
Hi @imnitishng sorry for the late reply! The PR is in good shape, but some work is still needed.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+            rich_bugtask['assignee_joined'] = None
+            rich_bugtask['assignee_karma'] = None
+            rich_bugtask['assignee_time_zone'] = None
+
+        # Extract info related to bug
+        rich_bugtask.update(self.__extract_bug_info(data['bug_data']))
+
+        if len(data['activity_data']) > 1:
+            rich_bugtask['time_to_first_activity'] = \
+                get_time_diff_days(data['date_created'], data['activity_data'][1]['datechanged'])
```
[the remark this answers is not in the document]
imnitishng: The first entry in activity data is always activity entry of the date on which bug was created, so using the first entry would cause `time_to_first_attention` to be 0 always.
The second entry reflects the first activity bug had.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+        """ Return a Sorting Hat identity using launchpad user data """
+
+        identity = {
+            'username': None,
+            'name': None,
+            'email': None
+        }
+
+        item = item['data']
+        if isinstance(item, dict) and item[identity_field]:
```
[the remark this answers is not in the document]
imnitishng: Sure, done.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+            rich_bugtask['assignee_login'] = None
+            rich_bugtask['assignee_name'] = None
+            rich_bugtask['assignee_joined'] = None
+            rich_bugtask['assignee_karma'] = None
+            rich_bugtask['assignee_time_zone'] = None
+
+        # Extract info related to bug
+        rich_bugtask.update(self.__extract_bug_info(data['bug_data']))
+
+        if len(data['activity_data']) > 1:
```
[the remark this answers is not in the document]
imnitishng: Sure, 
```
    "launchpad-implement": {
        "launchpad": [
            "https://launchpad.net/launchpad",
            "https://launchpad.net/ubuntu/+source/python-defaults"
        ]
    }
```

```
[launchpad]
raw_index = launchpad_raw
enriched_index = launchpad_enriched
from-date = 2020-01-15
```
the `launchpad` repo from this date will get bugs with `acticity_data`.

> I would suggest to replace the field time_to_first_activity with time_to_first_attention.

Sure, done.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+        """Create enriched data for bugs"""
+
+        rich_bugtask = {}
+
+        # Time to
+        rich_bugtask["created_to_assigned"] = get_time_diff_days(data['date_created'], data['date_assigned'])
+        rich_bugtask['assigned_to_closed'] = get_time_diff_days(data['date_assigned'], data['date_closed'])
+        rich_bugtask["time_to_close_days"] = get_time_diff_days(data['date_created'], data['date_closed'])
+
+        rich_bugtask['time_to_close_days'] = \
```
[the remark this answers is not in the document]
imnitishng: I'm sorry, yes.
I did it while creating dashboards, forgot to remove it.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+        for rol in self.roles:
+            if rol in item['data']:
+                user = self.get_sh_identity(item, rol)
+                yield user
+
+    @metadata
+    def get_rich_item(self, item):
+
+        eitem = {}
+
```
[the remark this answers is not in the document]
imnitishng: Done.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,219 @@
...
+        # Time to
+        rich_bugtask["created_to_assigned"] = get_time_diff_days(data['date_created'], data['date_assigned'])
+        rich_bugtask['assigned_to_closed'] = get_time_diff_days(data['date_assigned'], data['date_closed'])
+        rich_bugtask["time_to_close_days"] = get_time_diff_days(data['date_created'], data['date_closed'])
+
+        rich_bugtask['time_to_close_days'] = \
+            get_time_diff_days(data['date_created'], data['date_closed'])
+
+        if not data['date_closed'] and not data['is_complete']:
```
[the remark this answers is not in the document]
imnitishng:
```
        if not data['date_closed']:
            rich_bugtask['time_open_days'] = \
                get_time_diff_days(data['date_created'], datetime_utcnow().replace(tzinfo=None))
            if not data['is_complete']:
                rich_bugtask['time_still_open_days'] = rich_bugtask['time_open_days']
        else:
            rich_bugtask['time_open_days'] = rich_bugtask['time_to_close_days']
```

So the idea is to add a field named `time_still_open_days` to the enriched item, for the issues that are not marked complete (`is_complete = 0`). This denotes the issues that are still open in the repo.
Whereas, `time_open_days` is for the issues that have been completed, this denotes the time in days the issue remained open.

For example - 
An issue which is still open in the repo for 983 days will have the following data -

```
is_complete = 0
time_open_days = 983
time_still_open_days = 983
```

for an issue which was opened and then was closed or solved in 10 days will have this -
```
is_complete = 1
time_open_days = 10
time_still_open_days = attribute does not exist for this issue
```

The reason to do so was to get dashboards working, there are visualizations which are based off of these attributes.

grimoire_elk/enriched/launchpad.py
imnitishng: I understand your point, but similar method was used in the github issue enricher
reference, 
https://github.com/chaoss/grimoirelab-elk/blob/8c5319e01138800957beb3ad834d3c1f6ea27eb8/grimoire_elk/enriched/github.py#L536

Do we need to think of some different way of implementing this?

grimoire_elk/enriched/launchpad.py
imnitishng: > time_to_close_days is the time to close an issue (from its creation to the closing action). It is applied to issues that have been closed.

Correct

> time_open_days is the time an issue is open (from its creation until today). It is applied to issues that are not closed.
>> It is applied to issues that are not closed.

No, the field `time_open_days` is present in both closed and open issues, 
for closed it means the time to close an issue (that is technically the time issue remained open).
and for open issues it means the time from its creation to today

However, open issues also have a new attribute `time_still_open_days` as explained above.

grimoire_elk/enriched/launchpad.py
imnitishng: Yes, sorry for the mistake, should we rename this one to `time_created_to_last_update` ?

grimoire_elk/enriched/launchpad.py
imnitishng: `time_to_last_update_days` calculated here denotes the time difference from the latest update the bug had (status change, new comment, new fix, etc) to now.

`time_to_last_update_days` at 211 denotes the time difference between the day bug was created to the most recent update it had. I have proposed a rename to this attribute because they have the same name and store different values.

grimoire_elk/enriched/launchpad.py
imnitishng: Okay, will edit it.

grimoire_elk/raw/launchpad.py
imnitishng: Perceval as of now only has 1 argument `distribution` ([ref](
https://github.com/chaoss/grimoirelab-perceval/blob/5b4f848989967a51a4e21be1a3f6277f009f16ad/perceval/backends/core/launchpad.py#L520))

So if we are dealing with packages those are also passed as `distribution` parameter to the [constructor](https://github.com/chaoss/grimoirelab-perceval/blob/5b4f848989967a51a4e21be1a3f6277f009f16ad/perceval/backends/core/launchpad.py#L77) from elk ([here](https://github.com/chaoss/grimoirelab-elk/blob/8c5319e01138800957beb3ad834d3c1f6ea27eb8/grimoire_elk/elk.py#L85)).

So because of this we need to deal with distributions and packages URLs in the same way.
I have tested this implementation on this package from ubuntu https://launchpad.net/ubuntu/+source/python-defaults, and it works fine.

Please let me know I am wrong or if we need to change some stuff in perceval to support this.

grimoire_elk/enriched/launchpad.py
imnitishng: Sure. :smile:

grimoire_elk/enriched/launchpad.py
imnitishng: Thanks for the extra info, I will keep that in mind.

grimoire_elk/enriched/launchpad.py
imnitishng: Sure, I will do it.

grimoire_elk/enriched/launchpad.py
imnitishng: Thanks.

imnitishng:
Hi @valeriocos, thank you for the review.
I have some queries, will update the PR after those discussions.

grimoire_elk/enriched/launchpad.py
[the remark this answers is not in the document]
valeriocos: ok, thanks for the clarification

grimoire_elk/enriched/launchpad.py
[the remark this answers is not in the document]
valeriocos: no worries!

grimoire_elk/enriched/launchpad.py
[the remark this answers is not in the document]
valeriocos: I understand that you want to differentiate the time an issue was left open depending on the state of the issue (open or closed).

`time_still_open_days` seems redundant, since you can get the same with:
```
if is_complete == 0:
  time_open_days = today - creation time
else:
  time_open_days = closed time - creation time
``` 

Note that for every execution, the issues that have been updated upstream are also updated in ElasticSearch. For instance, let's imagine that there is an issue X upstream with state open for 20 days. We execute GrimoireLab and the corresponding enriched issue will contain `is_complete = 0` and `time_open_days = 20`.

5 days later the issue is closed upstream. When executing GrimoireLab, the issue will be fetched again and the previous enriched issue will be replaced with `is_complete = 1` and `time_open_days = 25`. The replacement is automatically done based on the `uuid` of the enriched item generated.

grimoire_elk/enriched/launchpad.py
valeriocos: You're right! The way the time_open_days for open issues is wrong. If you agree, we can focus on this PR and target the error on GitHub later.

The scripted field should be similar to:
```
if (doc.containsKey('is_complete') && !doc['is_complete'].value) { return Duration.between(LocalDateTime.ofInstant(Instant.ofEpochMilli(doc['grimoire_creation_date'].value.millis), ZoneId.of('Z')), LocalDateTime.ofInstant(Instant.ofEpochMilli(new Date().getTime()), ZoneId.of('Z'))).toMinutes()/1440.0; } }
```

Scripted fields can be created in the index pattern tab within Kibiter
captura_386

This information could be useful to be added at https://github.com/chaoss/grimoirelab-sirmordred/blob/master/Getting-Started.md#how-to-

grimoire_elk/enriched/launchpad.py
valeriocos: I agree with you about `time_open_days` being present in both cases.

WRT `time_still_open_days` I answered in the comment you mention

grimoire_elk/enriched/launchpad.py
valeriocos: yes, sounds good, thanks!

grimoire_elk/enriched/launchpad.py
valeriocos: ok, this could be renamed `time_created_to_last_activity_days` or just `time_to_last_activity_days` or `time_to_last_activity`.

grimoire_elk/enriched/launchpad.py
valeriocos: thanks!

grimoire_elk/raw/launchpad.py
valeriocos: Thank you for the clarification, and good catch about the missing param in Perceval.

The fix in Perceval should be easy, the code below should be added before this line https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/launchpad.py#L519.
```
group.add_argument('--package', dest='package',
                                  help="Distribution package")
```

Then a test should be added https://github.com/chaoss/grimoirelab-perceval/blob/master/tests/test_launchpad.py#L1002

The version number of the backend should be updated to 0.9.0: https://github.com/chaoss/grimoirelab-perceval/blob/master/perceval/backends/core/launchpad.py#L66

Let me know if you want to prepare the fix, otherwise I can do it.

grimoire_elk/enriched/launchpad.py
valeriocos: You're welcome!

valeriocos:
Hi @imnitishng , thank you for the quick reply and comments!

I have answered your queries, let me know if I missed something.

grimoire_elk/raw/launchpad.py
imnitishng: Sure thanks I will do it.

grimoire_elk/enriched/launchpad.py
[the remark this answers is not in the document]
imnitishng: Thank you for the details.
I have incorporated the changes.

grimoire_elk/enriched/launchpad.py
imnitishng: I have created the field `time_open_days`, regarding this script, I will make updates on the dashboards PR soon.

grimoire_elk/enriched/launchpad.py
imnitishng: Done, thanks.

imnitishng:
Hi @valeriocos I have updated the PR, please have a look.

grimoire_elk/enriched/launchpad.py
```diff
@@ -0,0 +1,193 @@
...
+
+        rich_bugtask['activity_count'] = len(data['activity_data'])
+
+        return rich_bugtask
+
+    def __extract_bug_info(self, bug_data):
+
+        rich_bug_info = {}
+
+        copy_fields = ["latest_patch_uploaded", "security_related", "private", "users_affected_count",
```
valeriocos, line 182: For the description field we should have 2 fields:
- description of type keyword. To avoid ES immense terms exceptions, the size value should be limited to KEYWORD_MAX_LENGTH
- description_analyzed of type text

This separation is done to create tables (or other vizs) able to aggregate on the keyword types, and leverage on searches using the lucene features provided by ES (for text types).
Git and GitHub includes this logic, please have a look at:
- https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/git.py#L222
- https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/git.py#L62
and
- https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/github.py#L581
- https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/github.py#L580
- https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/enriched/github.py#L83

tests/test_launchpad.py
```diff
@@ -0,0 +1,317 @@
...
+
+        result = self._test_raw_to_enrich(projects=True)
+        self.assertEqual(result['raw'], 4)
+        self.assertEqual(result['enrich'], 4)
+
+        res = requests.get(self.es_con + "/" + self.enrich_index + "/_search", verify=False)
+        for eitem in res.json()['hits']['hits']:
+            self.assertEqual(eitem['_source']['project'], "Main")
+
+    def test_perceval_params(self):
```
valeriocos, line 277: This test should enhanced to cover URLs including package info.

valeriocos (commented):
Thank you @imnitishng for working on it. We are close to finalize the PR, please check the comments when you have time.

Please make sure that the schema is consistent with the final version of the enricher.

Thanks!

imnitishng:
> With the recent change on Perceval (chaoss/grimoirelab-perceval@317e7f8), issues on package distribution cannot be fetched
```

Collection for launchpad: starting...

    response.raise_for_status()
  File "/home/slimbook/Escritorio/sources/venv/lib/python3.6/site-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://api.launchpad.net/1.0/python-defaults?order_by=date_last_updated&omit_duplicates=false&status=New&status=Incomplete&status=Opinion&status=Invalid&status=Won%27t+Fix&status=Expired&status=Confirmed&status=Triaged&status=In+Progress&status=Fix+Committed&status=Fix+Released&status=Incomplete+%28with+response%29&status=Incomplete+%28without+response%29&ws.op=searchTasks&modified_since=1970-01-01T00%3A00%3A00%2B00%3A00
  2020-05-03 11:25:45,815 [launchpad] Done collection for https://launchpad.net/python-defaults
```

Hi @valeriocos, regarding this issue.
The implementation does work correctly for distribution packages but the caveat is we need to specify an extra parameter in the `setup.cfg` file.

```
[launchpad]
raw_index = launchpad_raw
enriched_index = launchpad_enriched
from-date = 2020-01-15
package = python-defaults
```

This will be a problem since now we need to change our projects file to add the URL to which this package belongs.
```
        "launchpad": [
            "https://launchpad.net/ubuntu"
        ],
```

Also, now we cannot run micro-mordred to fetch and enrich data from both packages and distributions in a single run which completely defies the purpose of `projects.json` file, since for every package we will have to change the cfg file.
As far as I think we will have to change the perceval backend again, I have this option in mind

- Remove the package parameter completely as everything was working fine with the last version of backend. And we did support packages as well as distributions.
So our `projects.json` file will be something like this
```
        "launchpad": [
            "https://launchpad.net/launchpad",
            "https://launchpad.net/ubuntu/+source/python-defaults",
        ],
```
able to fetch data from both distros and packages without any concern.

schema/launchpad.csv
```diff
@@ -0,0 +1,79 @@
+name,type,aggregatable,description
+activity_count,int,true,"Count of activities on the bug. "
+assigned_to_closed,float,true,"Time difference from bug created to bug closed (in days). "
+assignee_data_bot,boolean,true,"True if the given assignee is identified as a bot."
+assignee_data_domain,keyword,true,"Assignee domain from Email. "
+assignee_data_gender,keyword,true,"Assignee gender. "
```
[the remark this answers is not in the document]
imnitishng: These attributes you mentioned were always None (hence of no use), so I thought of not including them.
I will update it now.

schema/launchpad.csv
```diff
@@ -0,0 +1,79 @@
...
+bug_id,keyword,true,"Bug ID on Launchpad. "
+bug_name,keyword,true,"Nickname for the bug. "
+bug_target_name,keyword,true,"Target of the bug. "
+created_to_assigned,date,true,"Time difference from bug created to bug assigned (in days). "
+date_created,date,true,"Date the bug was created. "
+date_incomplete,date,true,"Date the bug was marked incomplete. "
+date_last_updated,date,true,"Date the bug was last updated. "
+date_left_new,date,true,"Date the bug left status 'new'. "
+date_triaged,date,true,"Date the bug was marked Triaged. "
+description,keyword,true,"Description of the bug. "
```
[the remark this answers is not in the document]
imnitishng: Sure. Thanks.

schema/launchpad.csv
```diff
@@ -0,0 +1,79 @@
...
+message_count,int,true,"Messages(comments) made on the bug. "
+metadata__enriched_on,date,true,"Date when the item was enriched."
+metadata__gelk_backend_name,keyword,true,"Name of the backend used to enrich information."
+metadata__gelk_version,keyword,true,"Version of the backend used to enrich information."
+metadata__timestamp,date,true,"Date when the item was stored in RAW index."
+metadata__updated_on,date,true,"Date when the item was updated in its original data source."
+origin,keyword,true,"Original URL where the room was retrieved from."
+owner_data_bot,boolean,true,"True if the given author is identified as a bot."
+owner_data_domain,keyword,true,"Owner domain from Email. "
+owner_data_gender,keyword,true,"Owner gender. "
```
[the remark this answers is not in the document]
imnitishng: same comment as above

tests/test_launchpad.py
imnitishng: I included it in the previous iteration.
Had to change it because of the new perceval implementation.
Will add it after packages discussion is finalized.

schema/launchpad.csv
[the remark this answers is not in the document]
valeriocos: Thanks!

schema/launchpad.csv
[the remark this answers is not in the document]
valeriocos: Thanks!

valeriocos:
Hi @imnitishng , thank you for having a look at this!

> This will be a problem since now we need to change our projects file to add the URL to which this package belongs.

You are right! We cannot pass this param via the setup.cfg, since the latter is not suitable to handle params for the single repositories.

ELK provides the method `get_perceval_params_from_url` to process the URLs in the projects.json and shape them in a format suitable to execute Perceval. A possible implementation for the method for Launchpad could be the one below:
```
    @classmethod
    def get_perceval_params_from_url(cls, url):
        params = []
        splits = url.split('/')
        if '+source' in url:
            params.append(splits[-3])
            params.append('--package')
            params.append(splits[-1])
        else:
            params.append(splits[-1])

        return params
```

Similar solutions are applied to other backends to pass optional params:
- https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/raw/stackexchange.py#L69
- https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/raw/meetup.py#L75

WDYT?

imnitishng:
@valeriocos, thank you. So stupid of me, I did not think it this way.
I have updated the PR, please have a look. Let me know if some more work is needed.

valeriocos (approved):
You're welcome @imnitishng ! No worries!

LGTM, thanks! good job!

PR #855, opened 2020-04-24
[ELK] Add poetry support

valeriocos:
This code adds support for poetry, thus the poetry.lock and pyproject.toml are
committed to the repository.

The pyproject.toml has been derived from https://github.com/chaoss/grimoirelab-elk/blob/master/requirements.txt


sduenas (changes requested):
LGTM but there's a minor change you have to add before we merge it.

sduenas (approved):
LGTM

PR #856, opened 2020-04-24
[ELK] remove pymysql dependency

valeriocos:
This code removes the pymysql dependency used in enriched/enrich.py and enriched/projects.py.

Requirements.txt and setup.py have been updated accordingly.


jjmerchante:
The code has changed since this PR was created. We are now using Poetry and the code that this PR updates was removed.

I created a new PR that replaces this one but with the current source code: https://github.com/chaoss/grimoirelab-elk/pull/1120

PR #857, opened 2020-04-25
[utils] Remove get_git_repos

valeriocos:
This code removes the script get_git_repos from the folder utils, since it isn't used anywhere in the codebase.


zhquan (approved):
LGTM

PR #858, opened 2020-04-25
[grimoire_elk] Replace datetime/dateutil calls with toolkit

valeriocos:
This code replaces the calls to the datetime and dateutil calls to the equivalent ones provided by toolkit. This is needed to centralize the possible point of failures to a single library.


zhquan (approved):
LGTM

PR #859, opened 2020-04-25
[utils] Remove unused function get_time_diff_days

valeriocos:
This code removes the function `get_time_diff_days`, which isn't used in the codebase.


zhquan (approved):
Resolve the conflict

zhquan:
LGTM thanks @valeriocos

PR #861, opened 2020-04-25
[enriched] Remove projects.py

valeriocos:
This code removes the file projects.py, which isn't used in the rest of the codebase.


zhquan (approved):
LGTM

PR #864, opened 2020-04-27
[Enrich] Add `copy_raw_fields()` method and refactor enrichers

imnitishng:
This PR removes the following redundant code snippet from enrichers
```python
for f in self.RAW_FIELDS_COPY:
    if f in item:
        eitem[f] = item[f]
    else:
        eitem[f] = None
```
and replaces with new method `copy_raw_fields()`
Based on the discussion here https://github.com/chaoss/grimoirelab-elk/pull/851#discussion_r415019209

### DIFF
--- a/tests/test_crates.py
+++ b/tests/test_crates.py
@@ -94,6 +94,20 @@ class TestCrates(TestBaseBackend):
         self.assertEqual(result['raw'], 4)
         self.assertEqual(result['enrich'], 4)
 
+    def test_copy_raw_fields(self):
+        """Test copied raw fields"""
+
+        self._test_raw_to_enrich()
+        enrich_backend = self.connectors[self.connector][2]()
+
+        for item in self.items:
+            eitem = enrich_backend.get_rich_item(item)
+            for attribute in enrich_backend.RAW_FIELDS_COPY:
+                if attribute in item:
+                    self.assertEqual(item[attribute], eitem[attribute])
+                else:
+                    self.assertIsNone(eitem[attribute])
+
     def test_refresh_identities(self):
         """Test refresh identities"""
 
--- a/grimoire_elk/enriched/askbot.py
+++ b/grimoire_elk/enriched/askbot.py
@@ -121,11 +121,7 @@ class AskbotEnrich(Enrich):
     def get_rich_item(self, item):
         eitem = {}
 
-        for f in self.RAW_FIELDS_COPY:
-            if f in item:
-                eitem[f] = item[f]
-            else:
-                eitem[f] = None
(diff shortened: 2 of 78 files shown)


valeriocos (commented):
Thank you @imnitishng for the PR.

Can you add a test to https://github.com/chaoss/grimoirelab-elk/blob/master/tests/test_enrich.py to check that the method `copy_raw_fields` works fine

Probably some code could be added also here: https://github.com/chaoss/grimoirelab-elk/blob/master/tests/base.py#L197 to check that the default fields are copied to each enriched document.

The code could be as the one below:
```
        url = self.es_con + "/" + self.enrich_index + "/_search"
        response = enrich_backend.requests.get(url, verify=False).json()

        for hit in response['hits']['hits']:
            source = hit['_source']
            ...check that the default fields have been copied
```

WDYT?

imnitishng:
Sure, will update the PR.

valeriocos:
Thanks @imnitishng

imnitishng:
Hi @valeriocos, I have added the tests here https://github.com/chaoss/grimoirelab-elk/blob/master/tests/test_enrich.py .
But I wonder if adding tests at https://github.com/chaoss/grimoirelab-elk/blob/master/tests/base.py#L197 might not be efficient, since in every enricher `_test_raw_to_enrich()` is called multiple times, causing the unnecessary overhead for checking raw fields for every test item multiple times for every backend.
I think adding a new `test_raw_fields()` in every backend would be a better option for a new PR.

WDYT?

valeriocos:
Good idea @inishchith !

Ok for the new test (the name could be `test_copy_raw_fields`, since we are testing that some fields are copied from the raw index to the enriched index).

I would add this new test in this PR (as a separated commit) to make sure that ELK is working properly after this modification. WDYT?

imnitishng:
I have updated the tests @valeriocos, please have a look whenever you're free.

tests/test_crates.py
```diff
@@ -94,6 +94,20 @@ def test_raw_to_enrich_projects(self):
...
+        self._test_raw_to_enrich()
+        enrich_backend = self.connectors[self.connector][2]()
+
+        for item in self.items:
+            eitem = enrich_backend.get_rich_item(item)
+            for attribute in enrich_backend.RAW_FIELDS_COPY:
+                if attribute in item:
+                    self.assertEqual(item[attribute], eitem[attribute])
+                else:
+                    self.assertEqual(eitem[attribute], None)
```
valeriocos, line 109: this should be assertIsNone

valeriocos (commented):
Thank you @imnitishng for the PR. I left a minor comment

valeriocos (approved):
LGTM, thanks @imnitishng

PR #871, opened 2020-05-06
[enriched-cocom] Propagate commit parents to enriched index

valeriocos:
This code propogates the `parents` of the commit to the enriched documents. This allows to build scripts to navigate the commit graph.

Tests have been updated accordingly.


dlumbrer:
LGMT, thank you @valeriocos

PR #872, opened 2020-05-08
[enriched-mbox-study-kip] Fix calculation for inactive days

valeriocos:
This code fixes the way the attributes kip_discuss_inactive_days and kip_voting_inactive_days are calculated by not removing the tzinfo information. This change is needed since the function
`get_time_diff_days` needs both dates to be offset-aware.


zhquan (approved):
LGTM

PR #873, opened 2020-05-11
Handle Git private repos in ELK

valeriocos:
This code allows to handle Git private repos, which credentials are passed via the projects.json. To achieve this, the method `anonymize_url` in grimoire_elk/elastic.py has been removed and an equivalent function has been added to enriched/utils. This change is needed to ease the use of the
anonymize url feature across ELK.

The Git raw connector has been modified to include the method `_fix_item`, which takes care of removing the credentials that may appear in the origin. This change is needed to prevent storing in the raw index the credentials to access the Git private repos.

The method `get_repository_filter` in the enriched/utils has been modified to omit the credentials that may appear on the repo origin when creating the filter to get the data from the corresponding
raw index.

The study areas of code and git branches have been modified to omit the credentials in the repo urls to process. This change requires to also improve the way the attribute `project` is assigned to a repository. In the case of a private repo, the corresponding data source in the projects.json is accessed, each repo is anonymized and the output is compared with current repo (which value
is stored in the raw/enrich indexes).

Tests have been updated accordingly.

### DIFF
--- a/grimoire_elk/elastic.py
+++ b/grimoire_elk/elastic.py
@@ -21,7 +21,6 @@
 
 import json
 import logging
-import re
 import sys
 from time import time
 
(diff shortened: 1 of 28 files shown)


zhquan (approved):
LGTM

PR #878, opened 2020-05-19
[enriched-cocom] Include files in revision

valeriocos:
This code includes the attribute `files_at_commit`, which contains all files present in a given revision.


dlumbrer:
LGTM, thanks!!

PR #884, opened 2020-05-28
[enriched-git] Sync AOC and Git raw index

valeriocos:
This code allows to delete the documents corresponding to commits that don't exist anymore in upstream. This is achieved by calculating the difference between the commit hashes in the AOC and raw indexes, the obtained hashes are deleted from the AOC index.

This approach is similar to the one implemented in the method `update_items`, which allows to sync the Git raw and enriched indexes with the local clone.

This PR also extends the support for private git repositories when deleting commits from the git indexes (raw, enrich and aoc).


zhquan (approved):
LGTM

Issue #880, opened 2020-05-23
[Proposal] Plugin mechanism for ELK


heming6666:
## Why we need it?

There are more and more proposals about adding support for some backends that are not covered by Grimoirelab:

- Add Gitee backend support to GrimoireLab ELK: #835
- Implement Zulip backend to Grimoirelab ELK: #842 
- Add rocketchat backend support to ELK: #876

and so on.

**For `Perceval`**, some discussions were raised whether these backends should be integrated to the core of `perceval`:

- Add Airtable backend: https://github.com/chaoss/grimoirelab-perceval/pull/640
- Add support for Trello: https://github.com/chaoss/grimoirelab-perceval/issues/664

And currently we can have separate repositories for these Perceval-related backends, e.g. [chaoss/grimoirelab-perceval-mozilla](https://github.com/chaoss/grimoirelab-perceval-mozilla) and [WillemJiang/grimoirelab-perceval-gitee](https://github.com/WillemJiang/grimoirelab-perceval-gitee/). And they can be also imported by [`ELK`](https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/utils.py) in this way:
```python
from perceval.backends.core.git import Git, GitCommand
from perceval.backends.mozilla.crates import Crates, CratesCommand
from perceval.backends.opnfv.functest import Functest, FunctestCommand
```

However, **for `grimoirelab-elk`**, we currently don't provide such a mechanism to allow people to implement their beckends in a separate repository( `grimoirelab-elk-xxx`), since all of the connector meta data should be decleared in [`grimoire_elk/utils.py`](https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/utils.py).

If we can support plugin mechanism for `ELK`, then people can integrate the third party backends with the latest core easily. 
 
## How it could be?

Here is the [demo repository](https://github.com/heming6666/python-plugins-test) I write for testing plugin mechanism of python. 

In that case, `grimoirelab-elk-xxx` would act as a plugin and can be loaded or used by the core of `ELK`:

- For the plugins, they need to export the `connectors` (e.g. `Gitee`, `GiteeOcean`, `GiteeEnrich` and `GiteeCommand`) and use [`entry_points`](https://packaging.python.org/guides/creating-and-discovering-plugins/#using-package-metadata) to register themselves for discovery. 
[...]

PR #885, opened 2020-05-31
[utils] Use entry points to support plugin mechanism

heming6666:
This PR aims to support plugin mechanism for ELK. Related with #880 .

This code uses `pkg_resources.iter_entry_points()` to discover the registered entry point `grimoire_elk` and then load the `connectors` of external backends : 
```python
ENTRY_POINT_NAME = "grimoire_elk"

def get_connectors():
    # The default connectors
    connectors = {}
    for entry_point in pkg_resources.iter_entry_points(ENTRY_POINT_NAME):
        connectors.update(entry_point.load()())
    return connectors
```

Maybe we need pick up anothe suitable definition for `ENTRY_POINT_NAME` instead of `grimoire_elk`, I guess.

I will write some documentation for external packages about how to register themselves if necessary. 

Please @valeriocos @sduenas @WillemJiang help look into this and any feedback or suggestions are appreciated.

Signed-off-by: LinHaiming <lhming23@outlook.com>

### DIFF
--- a/grimoire_elk/utils.py
+++ b/grimoire_elk/utils.py
@@ -25,6 +25,8 @@ import sys
 
 import requests
 
+import pkg_resources
+
 from grimoire_elk.errors import ElasticError
 from grimoire_elk.elastic import ElasticSearch
 # Connectors for Graal
@@ -170,6 +172,7 @@ logger = logging.getLogger(__name__)
 
 kibiter_version = None
 
+ENTRY_POINT_NAME = "grimoire_elk"
 
 def get_connector_from_name(name):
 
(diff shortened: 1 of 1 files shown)


grimoire_elk/utils.py
```diff
@@ -266,6 +269,11 @@ def get_connectors():
             "twitter": [Twitter, TwitterOcean, TwitterEnrich, TwitterCommand]
             }  # Will come from Registry
 
+    for entry_point in pkg_resources.iter_entry_points(ENTRY_POINT_NAME):
+        connectors.update(entry_point.load()())
```
WillemJiang, line 273: We need to warn the user if there are some conflicts between the connectors are defined in the utils and the connectors which are loaded from entry_points.

heming6666: Thank you @WillemJiang very much. I have updated it. Please have a look again. 😃

eyehwan:
Hi @valeriocos, @sduenas and @vchrombie, 

Could you help review the code? It has been discussed in  #880, as the plugin proposal. 

As we discussed in https://github.com/chaoss/grimoirelab/issues/439, I have created a seperated repo [grimoirelab-gitee](https://github.com/grimoirelab-gitee) to include grimoirelab-gitee solution, that also depend on this plugin solution. 

Thanks.

jjmerchante:
This was implemented in https://github.com/chaoss/grimoirelab-elk/pull/1062

PR #886, opened 2020-06-02
[enriched-confluence] Extract username from publicName attr

valeriocos:
This code allows to extract the username information from the attribute `publicName`. This change is needed since some confluence instances don't use the attribute `username`

Tests have been added accordingly


zhquan (approved):
LGTM

PR #888, opened 2020-06-08
[elastic] Support for filter no-collection on multi params URLs

valeriocos:
This code allows to support the filter no-collection for URLs composed of multiples params, for instance:
```
"https://wiki-archive.opendaylight.org https://wiki-archive.opendaylight.org/view"
```


zhquan (approved):
LGTM

PR #889, opened 2020-06-10
[enriched-githubql] Add label attribute to closed events

valeriocos:
This code adds the attribute `label` which contains the issue labels to the closed events. This change is needed to allow calculations between label-related and closed events using a common attribute.


zhquan (approved):
LGTM

PR #893, opened 2020-06-18
[enriched-githubql] Add duration analysis study

valeriocos:
The purpose of this study is to calculate the duration between two GitHub events.

The study requires a start event type (e.g., UnlabeledEvent or MovedColumnsInProjectEvent), which is used to retrieve for each issue all events of that type.

For each issue event obtained, the first previous event of one of the types defined at `fltr_event_types` is returned, and used to calculate the duration (in days) between the two events. Optionally, an additional filter can be defined to retain the events that share a given property (e.g., a specific label, the name of project board).

Finally, the duration and the previous event uuid are added to the start event via the attributes
`duration_from_previous_event` and `previous_event_uuid`.

This study is executed in a incremental way, thus only the start events that don't include the attribute
`duration_from_previous_event` are retrieved and processed.

The examples below show how to activate the study by modifying the setup.cfg. The first example calculates the duration between Unlabeled and Labeled events per label. The second example
calculates the duration between the MovedColumnsInProject and AddedToProject events per column in each board.

```
[githubql]
...
studies = [enrich_duration_analysis:label, enrich_duration_analysis:kanban]

[enrich_duration_analysis:kanban]
start_event_type = MovedColumnsInProjectEvent
fltr_attr = board_name
target_attr = board_column
fltr_event_types = [MovedColumnsInProjectEvent, AddedToProjectEvent]

[enrich_duration_analysis:label]
start_event_type = UnlabeledEvent
target_attr = label
fltr_attr = label
fltr_event_types = [LabeledEvent]
```

Schema of githubql has been updated.


zhquan (approved):
LGTM

PR #894, opened 2020-06-18
[grimoire_elk] Remove tzinfo for date comparison

valeriocos:
This code removes the tzinfo used to compare the `from_date` value with the default one (1970-01-01) to perform the incremental fetching.

This change is needed to fix a bug introduced at:
- https://github.com/chaoss/grimoirelab-elk/commit/e2994480c6e4c40c7736f911e4287660e2863f31#diff-a94b51d8c8a4c2e33e52cff8968a0b3cR158
- https://github.com/chaoss/grimoirelab-elk/commit/e2994480c6e4c40c7736f911e4287660e2863f31#diff-928a29f9b639dcb91b5081e292e0d1c3R170

The bug consists in not removing the tzinfo from the value obtained from the method `str_to_datetime`.

### DIFF
--- a/grimoire_elk/elk.py
+++ b/grimoire_elk/elk.py
@@ -154,7 +154,7 @@ def feed_backend(url, clean, fetch_archive, backend_name, backend_params,
             params['branches'] = branches
         if filter_classified:
             params['filter_classified'] = filter_classified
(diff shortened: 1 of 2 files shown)


zhquan (approved):
LGTM

PR #895, opened 2020-06-19
[enriched-mbox] Handle long body emails

valeriocos:
This code reduces the size of the body emails collected from mbox-related repos.


zhquan (approved):
LGTM

PR #900, opened 2020-06-23
[enriched-githubql] Add page size param to duration study

valeriocos:
This code allows to tune from the setup.cfg the number of events without duration attribute returned per page. This change is useful to adapt the resources used by this study to the enviroment where GrimoireLab is running.


zhquan (approved):
LGTM

PR #901, opened 2020-06-23
[enriched-git] Execute update_items when local repo commits are returned

valeriocos:
This code prevents to delete commits from the raw and enriched indexes when no commits are returned from the local Git repository. This change is needed to cover the cases when an error is thrown during the execution of `rev_list` or when the location of the local repo has changed.


zhquan (approved):
LGTM

PR #903, opened 2020-06-29
Fix wrong quotation

jonaschn:


valeriocos (approved):
LGTM, thanks @jonaschn

PR #904, opened 2020-06-29
[enriched-enrich] Set user-agent for geopy

valeriocos:
This code sets the user-agent used by the study enrich_geolocation. This change is needed to support
geopy 2.0.0, which requires the declaration of a custom user-agent.


zhquan (approved):
LGTM

PR #929, opened 2020-08-12
[enriched/gerrit] Add new fields `open` and `wip`

zhquan:
This code adds two extra fields to the enriched index:
- `open`: from `data.open` True or False
- `wip`: from `data.wip` True or False (If the field `data.wip`
does not exist is set to False)

Test updated accordingly.


sduenas (changes requested):
The patch looks good but you missed to add these two fields to `gerrit.csv` on the `schema` directory.

sduenas (approved):
LGTM

PR #930, opened 2020-08-14
[schema] Fix slack schema CSV file

mafesan:
There was one row with one additional column and another one missing the `aggregatable` column.


schema/slack.csv
```diff
@@ -11,7 +11,7 @@ author_user_name,keyword,true,"Author username from Sortinghat profile."
 author_uuid,keyword,true,"Author UUID from SortingHat."
 avatar,keyword,true,"Avatar associated with User profile."
 channel_created,date,true,"Date when channel was created in UNIX timestamp format."
-channel_id,keyword,long,true,"Channel Id of a Slack channel."
+channel_id,long,true,"Channel Id of a Slack channel."
```
alpgarcia, line 14: This field should be a string because it contains alpha-numerical characters:
https://api.slack.com/methods/conversations.info

alpgarcia (changes requested):

schema/slack.csv
mafesan: Oops! My bad, sorry. Thank you for the heads-up!

alpgarcia (approved):
LGTM, thanks!

PR #938, opened 2020-09-24
[enriched/bugzilla] Add new field `whiteboard`

zhquan:
This code adds a new field to the enriched index that gives us
information about `whiteboard`.


sduenas (changes requested):
The PR looks ok but you need to update the CVS file with this information. I also think we should also add the same field for `bugzillarest` so we can have similar indices.

grimoire_elk/enriched/bugzillarest.py
```diff
@@ -156,6 +157,9 @@ def get_rich_item(self, item):
             if 'changes' in history:
                 eitem['changes'] += len(history['changes'])
 
+        if issue['whiteboard'] != "":
```
sduenas, line 160: Is this field in the item always? Not sure if that's going to create a `KeyError`.

zhquan: In Bugzillarest the field is always there but not in Bugzilla. So to have the same behavior:

Option 1:
- Bugzilla: Add the field `whiteboard` only when the field does exist on the raw index.
- Bugzillarest: Add the field `whiteboard` only when the field is not empty("") on the raw index

Option 2:
- Bugzilla: Add the field `whiteboard` always and when the field doesn't exist set as empty("").
- Bugzillarest: Add the field `whiteboard` always

I choose the option 1

sduenas (approved):
LGTM

PR #939, opened 2020-09-25
[enriched/git] `enrich_git_branches` skip --filter-no-collection repo

zhquan:
This code skips the study when `--filter-no-collection` is present
due to that repo was removed from upstream or we don't have
permission.

Test added accordingly.


sduenas (approved):
LGTM

PR #940, opened 2020-10-06
[weblate] New enricher to handle changes

zhquan:
This code proposes a new weblate enricher to handle changes
including their units and author information.

Demography study included.

The new enricher is added to the list of connetors and can be
defined in the mordred setup.cfg as follows:
```
[weblate]
raw_index = weblate_chaoss_raw
enriched_index = weblate_chaoss_enriched
api-token= ...
no-archive = true
sleep-for-rate = true
studies = [enrich_demography:weblate]

[enrich_demography:weblate]
```

projects.json
```
{
  "chaoss": {
    "weblate": [
      "my.weblate.org"
    ]
  }
}
```

Tests added accordingly.

### DIFF
--- a/grimoire_elk/enriched/weblate.py
+++ b/grimoire_elk/enriched/weblate.py
@@ -1,0 +1,177 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2015-2020 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#   Quan Zhou <quan@bitergia.com>
+#
+
+
+import logging
+
+from grimoirelab_toolkit.datetime import str_to_datetime
+
+from .enrich import Enrich, metadata
+from ..elastic_mapping import Mapping as BaseMapping
+
+
+logger = logging.getLogger(__name__)
+
+
+class Mapping(BaseMapping):
+
+    @staticmethod
+    def get_elastic_mappings(es_major):
+        """Get Elasticsearch mapping.
+
+        :param es_major: major version of Elasticsearch, as string
+        :returns: dictionary with a key, 'items', with the mapping
+        """
+
+        mapping = """
+        {
+            "properties": {
+                "text_analyzed": {
+                  "type": "text",
+                  "fielddata": true,
+                  "index": true
+                }
+           }
+        } """
+
+        return {"items": mapping}
+
+
+class WeblateEnrich(Enrich):
+
+    mapping = Mapping
+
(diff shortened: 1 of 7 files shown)


grimoire_elk/enriched/weblate.py
```diff
@@ -0,0 +1,161 @@
...
+        if change['user_data']:
+            user_data = change['user_data']
+            for data in user_data:
+                eitem['user_' + data] = user_data[data]
+
+        eitem['unit_api_url'] = change['unit']
+        if change['unit_data']:
+            user_data = change['unit_data']
+            for data in user_data:
+                eitem['unit_' + data] = user_data[data]
```
mafesan, line 129: I think there are some boolean fields here (`has_comment`, `has_failing_check` and `has_suggestion`) which could be converted to 0/1 values as we do in other similar fields from other enrichers.

Example:
https://github.com/chaoss/grimoirelab-elk/blob/cfbac57c84df8367fa4c914eb17b95e78686f99b/grimoire_elk/enriched/git.py#L375 

The reason for this conversion is to ease the aggregation of these fields. For instance, to have the number of translations (units) with comments per author, we would make a `sum` aggregation of the `has_comment` field.

mafesan, line 129: The fields `unit_content_hash` and `unit_id_hash` are of type `number` and they should be of type `string` according to Weblate doc (https://docs.weblate.org/en/latest/api.html#get--api-units-(int-id)-), please check how these fields are handled (maybe is something to fix from the Perceval backend).

grimoire_elk/raw/weblate.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class Mapping(BaseMapping):
+
+    @staticmethod
+    def get_elastic_mappings(es_major):
+        """Get Elasticsearch mapping.
+
+        :param es_major: major version of Elasticsearch, as string
+        :returns:        dictionary with a key, 'items', with the mapping
```
mafesan, line 33: Minor comment: please remove the extra spaces.

mafesan (changes requested):
@zhquan Overall it LGTM, but please check my comments. Thank you!

grimoire_elk/enriched/weblate.py
zhquan: But the API returns as `number` instead of `string` see https://translations.documentfoundation.org/api/units/24794381/

mafesan: Ok, thank you for checking it

requirements.txt
```diff
@@ -15,4 +15,5 @@ statsmodels >= 0.9.0
 -e git+https://github.com/chaoss/grimoirelab-perceval-mozilla/#egg=grimoirelab-perceval-mozilla
 -e git+https://github.com/chaoss/grimoirelab-perceval-opnfv/#egg=grimoirelab-perceval-opnfv
 -e git+https://github.com/chaoss/grimoirelab-perceval-puppet/#egg=grimoirelab-perceval-puppet
--e git+https://github.com/Bitergia/grimoirelab-perceval-finos/#egg=grimoirelab-perceval-finos
\ No newline at end of file
+-e git+https://github.com/Bitergia/grimoirelab-perceval-finos/#egg=grimoirelab-perceval-finos
+-e git+https://github.com/Bitergia/grimoirelab-perceval-weblate/#egg=grimoirelab-perceval-weblate
```
sduenas, line 19: This backend was moved to `chaoss`. Can you update the link?

sduenas (changes requested):
Just a minor comment. Update the weblate repository links.

requirements.txt
sduenas: Can you also rebase the PR?

sduenas (commented):
LGTM

PR #941, opened 2020-10-07
[enriched/githubql] Update references in related indexes

mafesan:
The cross reference study now accepts a list of related ES aliases or indices to be updated with the obtained references
between GitHub issues and Pull Requests, for all the elements identified by a given `issue_url`.

E.g., `github_issues` alias, which points to the corresponding enriched indexes from GitHub issues. Note that if the referenced
aliases or indexes do not contain the `issue_url` field, the affected elements can't be updated.

This is an example about how the `setup.cfg` has to be updated:

```
[githubql]
...
studies = [..., enrich_reference_analysis]

[enrich_reference_analysis]]
aliases_update = [github_issues, github2_issues, github2_pull_requests]
```

The affected index patters have been updated in this PR: https://github.com/chaoss/grimoirelab-sigils/pull/476

**Note**: as soon as this PR and https://github.com/chaoss/grimoirelab-sigils/pull/476 get merged, the index pattern for `github_issues` has to be updated to include the field `issue_url`, after the changes from https://github.com/chaoss/grimoirelab-elk/pull/942.

PR #942, opened 2020-10-07
[enriched/github] Add field issue_url for issues and pull requests

mafesan:
This field is added for preserving consistency with the rest of the related GitHub enrichers (`github2`, `githubql`), as both
are using this field to identify the URL of an issue or a pull request.

The value of this field is the same as the existing `url` field.

This PR is related to PR https://github.com/chaoss/grimoirelab-elk/pull/941.

**Note**: the related index pattern `github_issues` still has to be updated with this field. This has to be done after merging the changes from https://github.com/chaoss/grimoirelab-sigils/pull/476.


sduenas (approved):
LGTM

PR #945, opened 2020-10-20
[gerrit] Add last-verified approval value field to enricher

mafesan:
The new field is `changeset_status_value_verified` and it is defined as the value of the last "Verified" approval made by someone else than the author for a given changeset, it can be `-2`, `-1`, `1` or `2`.

Before this change, the last approval value for a given changeset was computed only for approvals labelled as "Code-Review". Now this last approval value is calculated for approvals labelled as
"Verified".

The CSV file containing the Gerrit schema has been updated accordingly.


zhquan (approved):
LGTM

PR #949, opened 2020-11-12
[raw/slack] Handle fields in 'data.files'

zhquan:
This code allows to handle the fields in `data.files` to
avoid the error `Failed to insert data to ES`


sduenas (approved):
LGTM

PR #950, opened 2020-11-13
[enriched/slack] Fix enrich items without user data

zhquan:
This code allows enriching items without user data.


sduenas (approved):
LGTM

PR #959, opened 2021-03-04
[enriched/gitlab] Fix KeyError without `name` value

zhquan:
This change fixes when we enrich a merge request item without the
`name` field in `item['data']['author']`.


sduenas (approved):
LGTM

PR #960, opened 2021-03-08
[enrich-stackexchange] Handle deleted user in stackoverflow

jjmerchante:
This code allows to deal with deleted users in answers in StackOverflow, logging a warning.

The implementation for this PR is similar to the code used to handle deleted users in questions.

One example of an answer without owner is https://stackoverflow.com/a/339388/15335404


sduenas (approved):
LGTM

PR #961, opened 2021-03-08
[stackexchange] Make author anonymization optional for stackexchange

jjmerchante:
This code makes it optional to pseudo-anonymize the owner of question and answers. It creates a hash of the name and user_id and remove author_link and profile_image.

For enabling the anonymization, include the following parameter in the stackexchange section:
```
[stackexchange]
...
anonymize = true
```


sduenas (approved):
LGTM

PR #964, opened 2021-03-13
[enriched/twitter] Remove deprecated attributes from enriched item

Rashmi-K-A:
The data returned from Twitter has deprecated fields that are set to null. We can remove these fields from the enriched twitter object since they are always going to be returned as null and are not going to be of much use. You can cross-check the deprecated fields [here.](https://developer.twitter.com/en/docs/twitter-api/v1/data-dictionary/object-model/user)

### DIFF
--- a/grimoire_elk/enriched/twitter.py
+++ b/grimoire_elk/enriched/twitter.py
@@ -142,8 +142,8 @@ class TwitterEnrich(Enrich):
 
         # data fields to copy from user
         copy_fields = ["created_at", "description", "followers_count",
-                       "friends_count", "id_str", "lang", "location", "name",
-                       "url", "utc_offset", "verified"]
+                       "friends_count", "id_str", "location", "name",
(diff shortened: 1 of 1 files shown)


sduenas (commented):
The PR looks good but we just need a minor change in the commit. Would you mind to add as a description of the commit what you wrote on the description of the PR. So the commit message will look as:
```
[enriched/twitter] Remove deprecated attributes from enriched item

The data returned from Twitter has deprecated fields that are set to null. We can remove
these fields from the enriched twitter object since they are always going to be returned
as null and are not going to be of much use. See developer's documentation for more info:

https://developer.twitter.com/en/docs/twitter-api/v1/data-dictionary/object-model/user

Signed-off-by: Rashmi K A <k.a.rashmi04@gmail.com>
```

sduenas (approved):
LGTM

PR #967, opened 2021-03-25
[schema] Fix twitter schema

Rashmi-K-A:
The aggregatable column for the field `place.bounding_box.coordinates` needs to be set to true.


sduenas (changes requested):
PR looks ok but would you mind to create a clean PR with the necessary commit only? As a tip, create a new branch from `origin/master`, add your commit there and push that branch. Try not to create PRs from your master branch.

Rashmi-K-A:
sure, not sure how I missed that. I'll close this one and do that 👍

sduenas:
You could have used the same PR. You just need to push the changes with `--force` to update this PR with the new state of your branch.

PR #969, opened 2021-03-26
[tests] Add test for metadata__updated_on field

VSevagen:
test_elastic_ocean.py was missing coverage for line 80. This commit fixes that.

Signed-off-by: sevagenv <sevagenv@gmail.com>

### DIFF
--- a/tests/test_elastic_ocean.py
+++ b/tests/test_elastic_ocean.py
@@ -22,6 +22,7 @@
 import unittest
 
 from grimoire_elk.raw.elastic import ElasticOcean
+from perceval.backends.core.git import Git
 from grimoire_elk.errors import ELKError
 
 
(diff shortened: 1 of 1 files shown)


tests/test_elastic_ocean.py
```diff
@@ -22,11 +22,16 @@
...
 
 from grimoire_elk.raw.elastic import ElasticOcean
+from perceval.backends.core.git import Git
 from grimoire_elk.errors import ELKError
 
 
 class TestElasticOcean(unittest.TestCase):
 
+    @classmethod
+    def setUpClass(cls):
```
sduenas, line 32: Why do we need this?

VSevagen: I'm not sure. Was getting an error that perceval_backend was not initialized

VSevagen: I tried with just `self` instead of `self.perceval_backend`. That also fixed the coverage for that line. However it seemed strange so I went with the 2nd option.

sduenas: So maybe it's better if you created it in that test instead for the whole test class.

VSevagen: ok. yeah. That would be better.

sduenas (approved):
LGTM

tests/test_elastic_ocean.py
sduenas: Thanks. Before we merge it, would you mind to rebase your commit with master?

VSevagen: yeah sure

sduenas (approved):
LGTM

sduenas:
Merged. Thanks for your contribution.

PR #971, opened 2021-03-28
[tests/data] Improve telegram coverage

VSevagen:
Coverage was missing for line 133-134.
This commit fixes that.

Signed-off-by: sevagenv <sevagenv@gmail.com>

### DIFF
--- a/tests/test_telegram.py
+++ b/tests/test_telegram.py
@@ -43,8 +43,32 @@ class TestTelegram(TestBaseBackend):
         """Test whether JSON items are properly inserted into ES"""
 
         result = self._test_items_to_raw()
-        self.assertEqual(result['items'], 9)
-        self.assertEqual(result['raw'], 9)
+        self.assertEqual(result['items'], 10)
+        self.assertEqual(result['raw'], 10)
+
+    def test_reply_to_message_text(self):
+        """Test whether item has text field available"""
+
+        self._test_items_to_raw()
+        enrich_backend = self.connectors[self.connector][2]()
+
+        for item in self.items:
+            message = item['data']['message']
+            eitem = enrich_backend.get_rich_item(item)
+            if 'reply_to_message' in message:
+                if 'text' in message['reply_to_message']:
+                    self.assertEqual(message['reply_to_message']['text'], eitem['reply_to_message'])
+
+    def test_reply_to_message_sticker(self):
+        """Test whether item has sticker field available"""
+
+        self._test_items_to_raw()
+        enrich_backend = self.connectors[self.connector][2]()
+
+        item = self.items[9]
+        message = item['data']['message']
+        eitem = enrich_backend.get_rich_item(item)
+        self.assertEqual(message['reply_to_message']['sticker'], eitem['reply_to_message'])
 
     def test_raw_to_enrich(self):
         """Test whether the raw index is properly enriched"""
--- a/tests/data/telegram.json
+++ b/tests/data/telegram.json
@@ -333,5 +333,80 @@
     "timestamp": 1588071688.3739,
     "updated_on": 1588071683.0,
     "uuid": "31f3fdb5641680f47bd415f1299adc53f2d9c3f2"
+},
+{
+    "backend_name": "Telegram",
+    "backend_version": "0.11.1",
+    "category": "message",
+    "classified_fields_filtered": null,
+    "data": {
+        "message": {
+            "chat": {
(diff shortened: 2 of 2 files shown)


sduenas (commented):
How does it exactly improves the coverage? What cases are checked here that weren't before?

VSevagen:
> How does it exactly improves the coverage? What cases are checked here that weren't before?

the `elif` condition on line 133 of telegram.py was never covered as the test case for it was not present in the json file. I just added an additional object without the `text` field to trigger the `sticker` condition on line 133.

sduenas:
> > How does it exactly improves the coverage? What cases are checked here that weren't before?
> 
> the `elif` condition on line 133 of telegram.py was never covered as the test case for it was not present in the json file. I just added an additional object without the `text` field but with the `sticker` field to trigger the `sticker` condition on line 133.

In that case, I think it's better to create a pair of tests where it's checked whether the `reply_to_message` contains the sticker or the text. Covering the case doesn't mean the case is working properly.

VSevagen:
Sorry for the delay @sduenas 
I made a pair of tests to check for  `message` and `sticker`. `text` is not really available as it get maps to `message`.

tests/test_telegram.py
```diff
@@ -46,6 +46,28 @@ def test_items_to_raw(self):
...
+    def test_reply_to_message_sticker(self):
+        """Test whether item has sticker field available"""
+
+        self._test_items_to_raw()
+        enrich_backend = self.connectors[self.connector][2]()
+
+        for item in self.items:
+            eitem = enrich_backend.get_rich_item(item)
+            if 'reply_to_message' in eitem:
+                self.assertIn('sticker', eitem)
```
sduenas, line 69: Can you explain this better? It looks like it happens in all the cases that we have. It should only work on those where the sticker is a reply.

VSevagen: so it should be asserting in eitem[reply_to_message] right ?

sduenas: This is the code your are testing:
```
            if 'text' in message['reply_to_message']:
                eitem['reply_to_message'] = message['reply_to_message']['text']
            elif 'sticker' in message['reply_to_message']:
                eitem['reply_to_message'] = message['reply_to_message']['sticker']
```

So what you need to check is that the value that is in `text` is stored in `eitem['reply_to_message']` when the first condition applies. If it's the second condition, you have to check it has the sticker value on in.

VSevagen:
@sduenas I updated the PR according to what you said. Not really sure if that's what you meant but I think I got it this time

sduenas:
> @sduenas I updated the PR according to what you said. Not really sure if that's what you meant but I think I got it this time

Kinda of. What I meant is. We know what item have the `sticker` data on it, so check that for that specific item, the exact value (the exact string) is there. It would be something like:

```
item = self.items[x] <- change x by the index number of the item with the sticker
[...]
self.assertEqual(message['reply_to_message']['sticker'], "sticker value") <- change 'sticker value' for the real value
```

VSevagen:
@sduenas but we're checking for `sticker` in `reply_to_message`. Right now, i think there is no item that has it. There are items that has `sticker` but not `reply_to_message` and vice versa.  I was checking telegram.json.

sduenas:
> @sduenas but we're checking for `sticker` in `reply_to_message`. Right now, i think there is no item that has it. There are items that has `sticker` but not `reply_to_message` and vice versa. I was checking telegram.json.
> Should I create an item ?

Yes, that's it. If there are no item with a value, let's create it because otherwise we are testing nothing.

VSevagen:
@sduenas I have a doubt regarding the `sticker` value. I didn't know how the sticker format would be like so i checked one existing format in the same json file. I'm not seeing any value for `sticker`. Its mostly a collection of other fields and values.

sduenas (approved):
LGTM

sduenas:
> @sduenas I have a doubt regarding the `sticker` value. I didn't know how the sticker format would be like so i checked one existing format in the same json file. I'm not seeing any value for `sticker`. Its mostly a collection of other fields and values.
> So to see what was being asserted, I checked the value for `message[reply_to_message][sticker]`. That turned out to be the collection of fields in sticker. Am I confusing something ?
> 
> I also updated the PR with the new item.

I think we can leave it that way. The PR looks good to me.

sduenas:
Merged. Thanks for your contribution.

PR #979, opened 2021-04-28
[enriched/github] Fix KeyError when getting user info

zhquan:
Perceval return this dict when the user is no longer exist or the
token doesn't have permission.
```
{"organizations": []}
```

User parameters are set to 'None' on this change.

### DIFF
--- a/grimoire_elk/enriched/github.py
+++ b/grimoire_elk/enriched/github.py
@@ -51,6 +51,10 @@ from .github_study_evolution import (get_unique_repository_with_project_name,
 
 GITHUB = 'https://github.com/'
 
+# Perceval returns this dict when the user no longer exists
+# or the token has no permission
+NO_USER_INFO = {"organizations": []}
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


grimoire_elk/enriched/github.py
```diff
@@ -51,6 +51,10 @@
 
 GITHUB = 'https://github.com/'
 
+# Perceval return this dict when the user is no longer exist
+# or the token has no permission
```
sanacl, line 55: It should be

``` # Perceval returns this dict when the user no longer exists```

File: grimoire_elk/enriched/github.py [hunk 2]
```diff
@@ -548,7 +550,7 @@ def __get_rich_issue(self, item):
         rich_issue['user_login'] = issue['user']['login']
 
         user = issue.get('user_data', None)
-        if user is not None and user:
+        if user is not None and user and user != NO_USER_INFO:
```
sanacl, line 553: why don't you create a method to evaluate whether user is available or not? this way the code is more readable, don't you think?

canasdiaz (changes requested):
One grammatical fix and a **suggestion** to make code more readable. If you don't like the suggestion I have nothing to say, my python skills are not in a good shape ;)

canasdiaz (approved):

sanacl:
@zhquan LGTM

PR #982, opened 2021-05-17
[enriched/mattermost] Improve fields

zhquan:
This change includes useful fields:
- channel_delete_at
- channel_team_id
- file_*
- hashtags
- is_pinned
- is_reply
- message
- message_analyzed
- message_id
- parent_id
- position
- reactions_*
- roles

Remove obsolete fields:
- avatar
- is_admin
- is_owner
- is_primary_owner
- number_attachs
- profile_title
- text
- text_analyzed

Schema update accordingly.


ajaragz:
Tested with a local docker Mattermost instance https://docs.mattermost.com/install/docker-local-machine.html. From a channel with multiple posts and reactions.

All fields added were found in the enriched documents. None of the removed fields are in the enriched documents.

Regarding new fields that may be interesting I don't see much else. I see that there's no email in the enriched fields but with the `user_data_name` and `user_data_domain` that's coevered.
 
LGTM

PR #983, opened 2021-06-04
[schema] Update Kafka KIP schema

canasdiaz:
This patch extends the information schema we have for Kafka KIP
following the standard format which includes the fields "name",
"type", "aggregatable" and "description" for every field in
the enriched index.

Signed-off-by: Luis Cañas-Díaz <lcanas@bitergia.com>

### DIFF
--- a/schema/kafka.csv
+++ b/schema/kafka.csv
@@ -1,56 +1,60 @@
-name,type
-author_bot,boolean
-author_domain,keyword
-author_id,keyword
-author_name,keyword
-author_org_name,keyword
-author_uuid,keyword
-body_extract,keyword
-body,text
-Date,keyword
-email_date,date
-From_bot,boolean
-From_domain,keyword
-From_id,keyword
-From,keyword
-From_name,keyword
-From_org_name,keyword
-From_uuid,keyword
-grimoire_creation_date,date
-is_pipermail_message,long
-kip_binding,long
-kip_discuss_inactive_days,float
-kip_discuss_time_days,float
-kip_final_status,keyword
-kip_is_discuss,long
-kip_is_first_discuss,long
-kip_is_first_vote,long
-kip_is_last_discuss,long
-kip_is_last_vote,long
-kip_is_vote,long
-kip,long
-kip_result,long
-kip_start_end,keyword
-kip_status,keyword
-kip_type,keyword
-kip_vote,long
-kip_voting_inactive_days,float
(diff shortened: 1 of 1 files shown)


zhquan:
@sanacl 
I miss these fields:
- author_gender
- author_gender_acc
- From_gender
- From_gender_acc
- From_user_name

And where does the `repository_labels` field come from?

sanacl:
@zhquan oh, good catch! you are right about the missing fields. 


You can see the code which assign the `repository_labels` at:
* https://github.com/chaoss/grimoirelab-elk/blob/df179d753ae3196d0143fe836d073e652e99ac36/grimoire_elk/enriched/mbox.py#L185
* https://github.com/chaoss/grimoirelab-elk/blob/df179d753ae3196d0143fe836d073e652e99ac36/grimoire_elk/enriched/utils.py#L42

sanacl:
@zhquan it is "fun" that the `repository_labels` field is not visible from the index pattern menu but it is from the discover app. I guess this is because of the two different indexes that are linked by the alias (mbox and pipermail indexes)

zhquan:
@sanacl ok, then we have to update the index pattern to include `repository_labels`

sanacl:
@zhquan should I add to the documentation all the fields visible from the discovery app? I guess the answer is yes, right?

zhquan:
@sanacl I see that the following fields are in the [index pattern](https://github.com/chaoss/grimoirelab-sigils/blob/master/json/kafka-index-pattern.json):
- `From_user_name`
- `From_gender`
- `From_gender_acc`

Undefined fields (there may be more):
- ` From_multi_org_names`
- `author_multi_org_names`
- `mbox_author_domain`
- `metadata__filter_raw`
- `offset`
- `repository_labels`

We have two options:
1. In this PR update only the available fields on the [index pattern](https://github.com/chaoss/grimoirelab-sigils/blob/master/json/kafka-index-pattern.json). And update again when the index pattern is updated
2. First update the index pattern and then update all fields in this PR

sanacl:
@zhquan in order to avoid being blocked and start "delivering" something I've done what you suggested in option 1. I've only included the ones available in the index pattern.

zhquan (approved):
LGTM

zhquan:
@sanacl rebase it and I will merge it thanks.

sanacl:
@zhquan done

PR #991, opened 2021-07-14
[raw/mattermost] Fix mapping for 'data.props.meeting_id'

zhquan:
This change allows mapping the attribute
`data.props.meeting_id` to a text type to avoid the error
`failed to parse field [data.props.meeting_id] of type [long] ..`.


sduenas (approved):
LGTM

PR #995, opened 2021-08-09
[gitlab] Autodetect gitlab enterprise-url arg

alch-emi:
The `enterprise-url` parameter of the `gitlab` backend is currently left to be specified by the user in the config file.  This commit removes this burden from the user by automatically detecting the enterprise URL from the url passed to get_perceval_params_from_url()`.

This aims to make it easier for users to work with enterprise GitLab instances by removing the need to use the `enterprise-url` at all.

As far as I know, this does not introduce any backwards-compatibility issues.

### DIFF
--- a/grimoire_elk/raw/gitlab.py
+++ b/grimoire_elk/raw/gitlab.py
@@ -80,6 +80,12 @@ class GitLabOcean(ElasticOcean):
         tokens = url.split(' ')
         repo = tokens[0]
 
+        # This removes the last two components from the URL (user & project) leaving only
+        # the host and protocol
+        host = '/'.join(repo.split('/')[:-2])
+        if host != 'https://gitlab.com':
+            params.extend(("--enterprise-url", host))
+
         owner = repo.split('/')[-2]
         repository = repo.split('/')[-1]
 


grimoire_elk/raw/gitlab.py
```diff
@@ -80,6 +80,10 @@ def get_perceval_params_from_url(cls, url):
         tokens = url.split(' ')
         repo = tokens[0]
 
+        backend = '/'.join(repo.split('/')[:-2])
```
sduenas, line 83: I would add a comment to explain that this is done to remove the owner and the repository from the repo url. It's a bit confusing, in my opinion, for someone like me that's reading the code for the first time.

grimoire_elk/raw/gitlab.py
```diff
@@ -80,6 +80,10 @@ def get_perceval_params_from_url(cls, url):
...
+        if backend != 'https://gitlab.com':
```
sduenas, line 84: I would rename `backend` to `host`. I makes more clear what the variable contains. We usually use backend with other meaning.

sduenas (changes requested):
Please check my comments. The patch looks ok, though. Thanks for the contribution.

grimoire_elk/raw/gitlab.py
Alch-Emi: Sounds good!

sduenas (approved):
LGTM

PR #997, opened 2021-08-17
[enriched/github2] Use user_data fields to set identities values

sduenas:
Name and username fields on the identities were using incorrect values. In the case of 'name' it was using 'login' value and not the name under 'user_data'. For the username it was using a field that doesn't not exist.

This patch fixes the problem using 'name' and 'login' for the two fields.


zhquan (approved):
LGTM

PR #998, opened 2021-08-17
[mattermost] Allow parsing standard mattermost urls

alch-emi:
With the addition of https://github.com/chaoss/grimoirelab-perceval/pull/748 to grimoirelab-perceval, it is now possible to call perceval using mattermost team & channel names.  This means that it is possible to accept a standard mattermost URL in grimoirelab-elk.  This PR adds support for this ability, while maintaining support for "legacy" repos in the old format.

This also adds a degree of validation to URLs, since they are now matched by regex, and improperly formatted URLs will fail to match.  I think this could be an opportunity to add an informative error message linking to the docs, but I wasn't able to find a place where the URL format is documented to link to.

[`mattermost.py:125`][] has an unfilled (TODO) where I think it would be good to link documentation.  I'd be happy to fill it in, and also update whatever documentation is currently available, but I'm not sure where to look.

[`mattermost.py:125`]: https://github.com/chaoss/grimoirelab-elk/blob/b5edb50e82423b8fb790cf49491f9f3fed04e5f1/grimoire_elk/raw/mattermost.py#L125

### DIFF
--- a/grimoire_elk/raw/mattermost.py
+++ b/grimoire_elk/raw/mattermost.py
@@ -19,6 +19,7 @@
 #   Alvaro del Castillo San Felix <acs@bitergia.com>
 #
 
+import re
 from .elastic import ElasticOcean
 from ..elastic_mapping import Mapping as BaseMapping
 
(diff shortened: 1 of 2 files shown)


grimoire_elk/raw/mattermost.py
```diff
@@ -80,15 +81,63 @@ class MattermostOcean(ElasticOcean):
...
+    URL_REGEX = re.compile('^'  # Pattern starts from the begining of the string
+
+                           # Both legacy and standard URLs start with the base URL
+                           r'(?P<base>https?://[\w\.\-]+\.[\w\-]+)'
+
+                           # Now match one of the types of URLs
+                           r'(?:'
+
+                           # Pattern for a standard URL
+                           r'(/(?P<team>[a-z\-_\d]+)/channels/(?P<channel>[a-z\-_\d]+)/?)'
```
sduenas, line 93: Do team names have only lower case letters? Is the same case with channels and with channels ids?

grimoire_elk/raw/mattermost.py
```diff
@@ -80,15 +81,63 @@ class MattermostOcean(ElasticOcean):
...
+
+                           # OR
+                           r'|'
+
+                           # Pattern for a legacy URL
+                           r'(/? (?P<channel_id>[a-z\d]{26}))'
```
sduenas, line 99: Is there a white space between the `/` and the channel id on legacy URLs?

sduenas (commented):
Thanks for your PR. It looks OK but I have some comments about how valid URLs are formed. Check my comments below.

grimoire_elk/raw/mattermost.py
Alch-Emi: Channel and team names can have uppercase letters, but these are converted to lowercase letters in the URL

![img][1]

[1]: https://media.githubusercontent.com/media/Alch-Emi/Image-Hosting/main/20210818.mattermost-team-name-transformation.png "An example of how mattermost URLs are transformed in the way described above"

grimoire_elk/raw/mattermost.py
Alch-Emi: There is indeed!  If you look through [the old code][1], you'll see that it is parsed by simply splitting the "URL" along the space, and simply passing the first two elements as parameters to perceval, and even failed if there was no space in the url

[1]: https://github.com/chaoss/grimoirelab-elk/blob/master/grimoire_elk/raw/mattermost.py#L89

Alch-Emi: Legacy "URL"s aren't really URLs at all, nor do they show up in mattermost's actual interface ever, and I don't think they're particularly well documented either.  That's part of why I'm hoping to supersede them with the actual URL format.  Support for legacy URLs is only kept around for backwards compatibility

Alch-Emi:
Does anyone know if there's documentation for the URL formats that I could link to here?

grimoire_elk/raw/mattermost.py
sduenas: But if I use capital letters for the URL when I already know the channel, I guess that works, right? My point is that maybe is better to use w+ that covers capital letters too, in case that someone makes a mistake with that. Just wondering.

Alch-Emi: That's a good point.  I just pushed a commit that should fix that, and also adds in a test for it.  I actually just call `.lower()` on the URL before hand, so any stray capitals anywhere in the URL will be normalized to lowercase, since im p sure perceval would break if you tried to pass it a capital-containing mattermost channel name.

zhquan (approved):
LGTM. Please rebase it

sduenas:
> Before I rebase, I'd like to just make sure that that [the reference to the docs](https://github.com/Alch-Emi/grimoirelab-elk/blob/mattermost-by-url/grimoire_elk/raw/mattermost.py#L125) that I mentioned earlier is filled out, so that the error message with the TODO in it doesn't accidentally get committed. I still don't know of any place where the URLs are documented though. Can you point me somewhere I could find that?

I think we don't have any kind of documentation. What we have is the perceval backend and that's all. I made that backend long ago so I don't remember the details. However, if I remember well, URLs were set concatenating the host address and the channel id because there wasn't a real URL. What I'm sure is that there's no white space in between both. You can check the original commit and the tests about it: https://github.com/chaoss/grimoirelab-perceval/commit/3f04b1c7b35ef04454f2c1070e9b3b522f03a0ad

Alch-Emi:
Yeah, the space existed only in the ELK format.  I'm quite confident that's the current URL format that it expects, even before this commit adds a new form.  You can see this [in the current test data][1], although the parameters are transformed before they are passed to perceval.  It's unfortunate that there is no place that this is documented though.  I'll just remove the reference for now.

[1]: https://github.com/chaoss/grimoirelab-elk/blob/a91fca2bdfb4a15e96545bd151927baa1d4d03c9/tests/test_mattermost.py#L135

Alch-Emi:
Alright, just pushed the change, should be good to merge if everyone is still okay with it.

sduenas (approved):
LGTM

PR #1001, opened 2021-08-30
[enriched/github2] Add 'user_login' field

zhquan:
This code adds the field `user_login` for all enriched items
when the category is `issue` or `pull_request`.

Test updated accordingly.

### DIFF
--- a/grimoire_elk/enriched/github2.py
+++ b/grimoire_elk/enriched/github2.py
@@ -285,6 +285,9 @@ class GitHubEnrich2(Enrich):
             ecomment['is_github_{}'.format(ISSUE_COMMENT_TYPE)] = 1
             ecomment['is_github_comment'] = 1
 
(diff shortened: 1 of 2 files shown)


grimoire_elk/enriched/github2.py
```diff
@@ -449,7 +455,7 @@ def __get_rich_pull(self, item):
 
         user = pull_request.get('user_data', None)
         if user is not None and user:
-            rich_pr['user_name'] = user['name']
+            rich_pr['user_name'] = user['login']
```
sduenas, line 458: Why this change? I see that in this same file there are other references to 'user_name' that are not changed.

zhquan: I changed this by mistake

sduenas (approved):
LGTM

PR #1003, opened 2021-09-01
[enriched/github2] Fix TypeError NoneType

zhquan:
This code fixes when the comment belongs to a user that was deleted
and replaced by https://github.com/ghost. It also sets
'Deleted user' as 'author_name'.


sduenas (approved):
LGTM

PR #1004, opened 2021-09-06
[elk] Pass 'no_update' to pereceval

### LINKED ISSUE #996: `no_update` is not transmitted to the perceval backend.
From https://github.com/chaoss/grimoirelab-perceval/blob/8ed4fb18055b5cb0fe6fc67c80ab50b285e306df/perceval/backends/core/git.py#L375
```
        exgroup_fetch.add_argument('--no-update', dest='no_update',
```
we can set `no-update` to fetch all commits without updating the repository.

However, even if I set `no-update = true` in `setup.cfg` when I used sirmordred module, the parameter wasn't transmitted to the perceval backend.

I think the reason is that the `no-update` is hidden in `grimoire_elk.elk.feed_backend`.
https://github.com/chaoss/grimoirelab-elk/blob/00b346623d8f3f0b4de901d25a0ef28e79a02545/grimoire_elk/elk.py#L81

Here, `no-update` is true which we set in parsed_args.

https://github.com/chaoss/grimoirelab-elk/blob/00b346623d8f3f0b4de901d25a0ef28e79a02545/grimoire_elk/elk.py#L148-L160

Here, `no-update` is not transmitted when we call `ocean_backend.feed(**params)`.

And then, in the method `ocean_backend.feed`, we lost the `no-update` so that we would use the default value `false` for `no-update` in `Git.fetch` forever.

zhquan:
This code passes the `no_update` parameter set in the
`setup.cfg` to perceval to fetch items. Also, this parameter
cannot be used together with `latest_items`.

Fixes #996

### DIFF
--- a/grimoire_elk/elk.py
+++ b/grimoire_elk/elk.py
@@ -75,6 +75,7 @@ def feed_backend(url, clean, fetch_archive, backend_name, backend_params,
         branches = None
         latest_items = None
         filter_classified = None
+        no_update = None
 
         backend_cmd = klass(*backend_params)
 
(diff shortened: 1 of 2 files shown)


sduenas (approved):
LGTM

PR #1005, opened 2021-09-06
[raw/gitlab] Handle immense term for data.head_pipeline.yaml_errors

zhquan:
This code modifies the raw mapping index to handle immense terms
found in `data.head_pipeline.yaml_errors`.

Tests data updated accordingly.


sduenas (approved):
LGTM

PR #1006, opened 2021-09-15
[enriched/github2] Add ghost user to calculate first attention

zhquan:
This code takes into account the deleted (ghost) users to
calculate the first attention and the first response. Also, fix
the `TypeError: 'NoneType'` error when the attention was made by
the ghost user.


sduenas (approved):
LGTM

PR #1018, opened 2021-12-22
[enrich] Get main enrollment for *_org_name field

zhquan:
This code gets the main enrollment for `*_org_name` when
there are several enrollments.

If the enrollment contains :: the main one is the first part.

For example:
- Enrollment: Chaoss::Eng
- Main: Chaoss

If there is more than one, it will return the first one ordered
alphabetically.

Test added accordingly.

### DIFF
--- a/grimoire_elk/enriched/enrich.py
+++ b/grimoire_elk/enriched/enrich.py
@@ -710,6 +710,22 @@ class Enrich(ElasticItems):
 
         return enrolls
 
+    @staticmethod
+    def get_main_enrollments(enrollments):
+        """ Get the main enrollment given a list of enrollments.
+        If the enrollment contains :: the main one is the first part.
+
+        For example:
+        - Enrollment: Chaoss::Eng
+        - Main: Chaoss
+
+        If there is more than one, it will return ordered alphabetically.
+        """
+        main_orgs = list(map(lambda x: x.split("::")[0], enrollments))
+        main_orgs = sorted(list(set(main_orgs)))
+
+        return main_orgs
+
     def __get_item_sh_fields_empty(self, rol, undefined=False):
         """ Return a SH identity with all fields to empty_field """
         # If empty_field is None, the fields do not appear in index patterns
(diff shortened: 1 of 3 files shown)


grimoire_elk/enriched/enrich.py
```diff
@@ -799,10 +818,13 @@ def get_item_sh_fields(self, identity=None, item_date=None, sh_id=None,
             eitem_sh[rol + "_gender"] = self.unknown_gender
             eitem_sh[rol + "_gender_acc"] = 0
 
-        eitem_sh[rol + "_org_name"] = self.get_enrollment(eitem_sh[rol + "_uuid"], item_date)
         eitem_sh[rol + "_bot"] = self.is_bot(eitem_sh[rol + '_uuid'])
 
-        eitem_sh[rol + MULTI_ORG_NAMES] = self.get_multi_enrollment(eitem_sh[rol + "_uuid"], item_date)
+        multi_enrolls = self.get_multi_enrollment(eitem_sh[rol + "_uuid"], item_date)
+        main_enrolls = self.get_main_enrollments(multi_enrolls)
+        all_enrolls = multi_enrolls + list(set(main_enrolls) - set(multi_enrolls))
```
sduenas, line 825: I don't get this. Shouldn't it be only:

```
list(set(main_enrolls) + set(multi_enrolls))
```
???

sduenas (changes requested):
Looks ok but we can improve it a bit using python functions.

grimoire_elk/enriched/enrich.py
zhquan:
```
all_enrolls = multi_enrolls + list(set(main_enrolls) - set(multi_enrolls))
````
We have:
- multi_enrolls = ["Bitergia::Eng", "Chaoss", "Bitergia"]
- main_enrolls = ["Bitergia", "Chaoss"]

We need to get the unique values between these two lists:
`all multi_enrolls` + all in `main_enrolls` but not in `multi_enrolls`
- all_enrolls = ['Bitergia::Eng', 'Chaoss', 'Bitergia']

sduenas: Yes, but if you have those into sets and you add both sets, you'll have the result you mention there.

sduenas (approved):
LGTM

PR #1036, opened 2022-02-28
Remove Python 3.6 support

jjmerchante:
This PR removes Python 3.6 support because the support ended on 23 Dec 2021.

Python 3.7 and 3.8 are now supported.

Related to https://github.com/chaoss/grimoirelab/issues/460


vchrombie:
The PR looks good to me, but can we add python 3.9 too to the CI configuration? :thinking:

jjmerchante:
Yes, I wanted to add 3.9 but the problem is that there are many dependencies pinned for a reason.

For example, `numpy` and `pandas` are pinned to a version that is not compatible with Python 3.9.

If I try to change the version of Pandas to <1.4 (latest version compatible with 3.7) it keeps installing an older version of Pandas because sortinghat and sigils have it pinned.

Since we have to play with the versions of all the packages, yesterday we talked that unless it is as simple as putting 3.9 and it works, for now keep 3.7 and 3.8 and we have two and a half years to include 3.9.

vchrombie:
Ok @jjmerchante, now it makes sense to me. I agree with keeping 3.7 and 3.8 for now. :+1:

zhquan (approved):
LGTM

PR #1049, opened 2022-05-10
[enriched/git] Fix 'is_git_commit_multi_author' field.

zhquan:
This fixes the `is_git_commit_multi_author` field when the
commit has more than one author and `pair-programming` is enabled.
This is important for calculating `pair_programming_*` values.

Also, set `git_uuid` correctly.


jjmerchante:
LGTM

PR #1050, opened 2022-05-13
Set SortingHat as a required package in grimoireelk

jjmerchante:
SortingHat was set as an extra dependency in elk, but actually it is required for running elk.

This PR updates the Poetry dependencies to make it a required package and updates the GitHub actions to remove Sortinghat as an extra dependency.

### DIFF
--- a/releases/unreleased/set-sortinghat-as-a-required-package-in-elk.yml
+++ b/releases/unreleased/set-sortinghat-as-a-required-package-in-elk.yml
@@ -1,0 +1,9 @@
+---
+title: SortingHat as a required package
+category: fixed
+author: Jose Javier Merchante <jjmerchante@bitergia.com>
+issue: null
+notes: >
+    SortingHat is required for running grimoireelk but
+    it was set as an extra dependency. From now on sortinghat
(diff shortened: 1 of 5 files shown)


sduenas:
Related to: https://github.com/chaoss/grimoirelab/issues/489

sduenas (changes requested):
LGTM but you should update the dependencies. They were updated in the last PR we merged. Thanks!

jjmerchante:
PR rebased and dependencies updated

releases/unreleased/set-sortinghat-as-a-required-package-in-elk.yml
```diff
@@ -0,0 +1,9 @@
+---
+title: Set SortingHat as a required package in elk
```
sduenas, line 2: I think this title is better for the changelog.

```suggestion
title: SortingHat as a required package
```

sduenas (changes requested):
LGTM but check my comment if it makes sense to you.

sduenas (approved):
LGTM

PR #1052, opened 2022-05-13
Add methods to add identities from Enrich class

### LINKED ISSUE #489: Break circular dependency between cereslib and grimoireelk
There's a circular dependency between `cereslib` and `grimoireelk` which makes impossible to update other package dependencies.

The root of this problem is that `cereslib` needs to update the information about identities in the database. To do so, it uses the connector with `sortinghat` that `elk` has. At the same time, `elk` uses `cereslib` to generate data events for a later use in the studies. This is clearly a code smell and we need to fix it.

After analyzing the problem, we think that `cereslib` doesn't really need to write in the database. This is something `elk` can do. We just need to generate an iterator of events and store identities' data when every event is processed.

These are the tasks needed to fix this issue:

 - [x] Break the code dependency between the two packages
 - [x] Set `sortinghat` as a required package in `elk`
 - [x] Remove the dependency of packages in `cereslib` with `elk`

sduenas:
The methods 'add_identities' and 'add_identity' have been added to 'Enrich' class to add new identities to a SortingHat
database. Libraries using elk won't need to use sortinghat library to store new identities while enriching items.

This helps to fix https://github.com/chaoss/grimoirelab/issues/489.

### DIFF
--- a/grimoire_elk/enriched/enrich.py
+++ b/grimoire_elk/enriched/enrich.py
@@ -1019,6 +1019,14 @@ class Enrich(ElasticItems):
         """ Get the SH identity uuid from the id """
         return SortingHat.get_uuid_from_id(self.sh_db, sh_id)
 
+    def add_sh_identities(self, identities):
+        SortingHat.add_identities(self.sh_db, identities,
+                                  self.get_connector_name())
+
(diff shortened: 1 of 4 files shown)


zhquan (approved):
LGTM

PR #1053, opened 2022-05-20
[elk] Skip sources with wrong arguments

zhquan:
When the source has a wrong argument, the current behavior is that
the process stops without any error messages.

This code will write an error message with the arguments
(anonymized) and skip the source so as not to stop the process.

Test added accordingly.

### DIFF
--- a/releases/unreleased/skip-sources-with-wrong-arguments.yml
+++ b/releases/unreleased/skip-sources-with-wrong-arguments.yml
@@ -1,0 +1,8 @@
+---
+title: Skip sources with wrong arguments
+category: fixed
+author: Quan Zhou <quan@bitergia.com>
+issue: null
+notes: >
+    Sources won't interrupt collection or enrichment processes when any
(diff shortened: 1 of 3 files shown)


jjmerchante:
LGTM

releases/unreleased/skip-sources-with-wrong-arguments.yml
```diff
@@ -0,0 +1,11 @@
...
+title: Skip sources with wrong arguments
+category: fixed
+author: Quan Zhou <quan@bitergia.com>
+issue: null
+notes: >
+    When the source has a wrong argument, the current behavior is that
+    the process stops without any error messages.
+
+    This feature will write an error message with the arguments
+    (anonymized) and skip the source so as not to stop the process.
```
sduenas, line 11:
```suggestion
    Sources won't interrupt enriching processes when any of their arguments 
    are invalid. Instead, they will be skipped.
```

sduenas (approved):
LGTM but please change the changelog entry

PR #1055, opened 2022-05-23
[elk] Add method to anonymize parameters

zhquan:
This code adds a new method to anonymize parameters in case you
want to write it in a log message.

All parameters after `--api-token` or `--backend-password` will
be replaced by `xxxxx` until the next parameter starts with `-`

Test added accordingly.


sduenas (changes requested):
LGTM but we can improve the changelog message. Please check it.

sduenas (approved):

PR #1056, opened 2022-05-23
Remove support of FINOS Meetings backend

sduenas:
The backend is archived in its original repository. We consider it as unmaintained software so
it shouldn't be part of the core of GrimoireLab.


vchrombie:
Related to https://github.com/chaoss/grimoirelab/issues/468

zhquan (approved):
LGTM

PR #1059, opened 2022-05-30
[enrich/demographics] Fix fetching authors with min and max dates

zhquan:
This code fixes fetching all authors with min and max dates.
Currently, ElasticSearch and OpenSearch use `composite` aggregation
to paginate all buckets.

The old query only returns 10000 items and also returns the
following warning message:

```
Deprecation: This aggregation creates too many buckets (10001) and
will throw an error in future versions. You should update the
[search.max_buckets] cluster setting or use the [composite]
aggregation to paginate all buckets in multiple requests.
```

Tests added acconrdingly.

### DIFF
--- a/grimoire_elk/enriched/enrich.py
+++ b/grimoire_elk/enriched/enrich.py
@@ -1879,29 +1879,13 @@ class Enrich(ElasticItems):
             In case there is no specific contribution type, by default all contributions will be considered.
         """
         # The first step is to find the current min and max date for all the authors
-        authors_min_max_data = {}
-
-        es_query = Enrich.authors_min_max_dates(date_field,
-                                                author_field=author_field,
-                                                contribution_type=contribution_type)
-        r = self.requests.post(self.elastic.index_url + "/_search",
-                               data=es_query, headers=HEADER_JSON,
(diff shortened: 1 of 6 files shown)


grimoire_elk/enriched/enrich.py
```diff
@@ -1927,6 +1913,66 @@ def run_demography(self, date_field, author_field, log_prefix, contribution_type
...
+            logger.error(ex)
+            return
+        aggregations_author = r.json()['aggregations']['author']
+        if not aggregations_author['buckets']:
+            return authors_min_max_data
+
+        for author in aggregations_author['buckets']:
+            authors_min_max_data[author['key'][author_field]] = author
+
+        after = aggregations_author['after_key'][author_field]
```
jjmerchante, line 1949: Just one minor comment:

This code is duplicated and in the future we might need to change it twice. Maybe a `while True` and a `break` in some specific points would improve it, what do you think?

jjmerchante (changes requested):

jjmerchante:
LGTM

jjmerchante:
LGTM

PR #1064, opened 2022-06-23
Update poetry builder version

sduenas:
The current version of the poetry builder was old and caused some problems when installing packages in developer mode. With this new version, it's possible to install ELK in developer or editable mode without using setuptools files (i.e. `setup.cfg` and `setup.py`).


vchrombie (approved):
LGTM!

Thanks for the PR @sduenas. Now the poetry configurations align across the tools.

PR #1078, opened 2022-09-26
[deps] Update PyMySQL to use the latest version

jjmerchante:
This PR removes the restriction to PyMySQL to use the latest available version. This is the same version as in SortingHat.


sduenas (approved):
LGTM

PR #1080, opened 2022-10-18
[enriched/gitlab] Fix unexpected keyword argument 'alias' in onion

zhquan:
This code fixes TypeError since the name of the aliases is no longer harcoded, now it is passed by parameter.

Error:
```
TypeError: enrich_onion() got an unexpected keyword argument 'alias'
```

Test added accordingly.


sduenas (approved):
LGTM

PR #1088, opened 2022-11-03
[enriched/weblate] Fix unexpected keyword argument 'alias' in demography

zhquan:
This code fixes TypeError in weblate since the name of the aliases is no longer harcoded, now it is passed by parameter.

Error:
```
TypeError: enrich_demography() got an unexpected keyword argument 'alias'
```

Test added accordingly.


jjmerchante:
LGTM

PR #1093, opened 2022-12-20
Fix mapping and bulk URL for ES 8.X and OS 2.X

Rafaeltheraven:
The way support for ES 7.X and OS 1.X was originally implemented meant that any major version that was not EXACTLY 7/1 would revert to old behavior. This is obviously wrong and simply changing this check to allow versions >= 7/1 seems to just work (at least in the context of using mordred to import git data from perceval).

I can not guarantee that everything works, and I know that at least kibash still needs some work as discussed here. But in any case, the check was wrong, and this PR should at least fix that.

:edit: Recreated this PR as I messed up what branch I used as the source for this PR and the old one got unclean.

### DIFF
--- a/grimoire_elk/elastic.py
+++ b/grimoire_elk/elastic.py
@@ -298,8 +298,8 @@ class ElasticSearch(object):
     def get_bulk_url(self):
         """Get the bulk URL endpoint"""
 
-        if (self.major == '7' and self.distribution == 'elasticsearch') or \
-           (self.major == '1' and self.distribution == 'opensearch'):
+        if (int(self.major) >= 7 and self.distribution == 'elasticsearch') or \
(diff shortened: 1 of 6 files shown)


jjmerchante:
Thanks for your patch. Could you rebase the PR (cause the tests are failing) and squash the commits?

I also see some unrelated code to the PR inside the `.github/workflows` directory.

Thank you

Rafaeltheraven:
My apologies, the project for which these fixes were originally made ended so I got distracted, I've cleaned up the PR now and squashed everything into 1 commit.

jjmerchante:
LGTM

Can you please add a changelog like in the other pull request?

zhquan:
Thank you very much for your contribution

PR #1095, opened 2023-01-11
[enrich] Include SortingHat parameters

jjmerchante:
This PR includes `sh_port`, `sh_path`, `sh_ssl` parameters to all the backends. These parameters are now mandatory in the configuration.

Tests updated accordingly.


sduenas (commented):
You mentioned this also fixes an error with the SortingHatClient but I don't see any specific change.

sduenas (changes requested):
I saw what you mentioned about the problem with the exception. I suggest to remove that piece of code and add a new PR with that change, because it's not related or include it in a new commit and change the desc of the PR.

sduenas (approved):
LGTM

PR #1099, opened 2023-01-16
Specify exact mappings for date fields in git documents

Rafaeltheraven:
Simple PR, just specifies that `AuthorDate` and `CommitDate` are date fields in ES. This allows us to later do better filtering such as being able to specify `AuthorDate <= {Some Date}`.

PR #1100, opened 2023-01-24
Define the date format for git documents exactly as the default for `git log`

Rafaeltheraven:
This is a patch for my earlier patch in #1099. 

Some background: my own testing revealed that the documentation for java date formatting isn't completely correct and that `dd` does not always accept a non-zero padded day-of-month. I've subsequently delved into the git source code (the documentation is wrong but that's another story) and discovered that the default date format for `git log` has a non-zero-padded day-of-month and also optional timezone information depending on the users locale.

Technically, the date format for a `git log` message could be anything, but it can generally be assumed to be the default. Not sure how the CHAOSS team wants to deal with this possibility. Personally I think anyone who changes the default date format for their git log can be expected to know what they are doing and to troubleshoot the issue, but maybe it's good to document this somewhere.

The `git log` default formatting code is found [here](https://github.com/git/git/blob/56c8fb1e95377900ec9d53c07886022af0a5d3c2/date.c#L266)


Rafaeltheraven:
@sduenas pinging you since you also looked at my previous PR.

sduenas (changes requested):
Sorry for being late on this. Last weeks we have been preparing the CHAOSSCon and FOSDEM and I completely missed this PR.
The PR looks ok but I would like you - if possible - to write a changelog entry so we can include this fix on the changelog notes we generate with every release. I can write the changelog by myself but I need write permissions on your branch to add a new commit with it.

For example: https://github.com/chaoss/grimoirelab-elk/blob/9672086cb0393185fb70d0757d127a7cb4cf1112/releases/unreleased/processed/sortinghat-database-changed.yml

Rafaeltheraven:
Sorry, got distracted as the project for which I originally wrote these patches reached it's end. Wrote a changelog as requested

zhquan:
Hi @Rafaeltheraven could you rebase your branch again and I will proceed to merge this PR

zhquan (approved):
LGTM

PR #1104, opened 2023-03-07
Fix typo in SortingHat API fields

jjmerchante:
This PR updates the name of the fields that are obtained from the SortingHat API. Previously the fields were `gender_acc` and `is_bot` but now are `genderAcc` and `isBot`.

Update tests accordingly.


sduenas (approved):
LGTM

PR #1110, opened 2023-07-07
Improve enrichment process performance

zhquan:
This commit improves the performance of the enrichment process.

- Reduce the number of requests to get the backend name (`_connecto_name`).
- Reduce the number of queries to get information from SortingHat.
- Removes an unused method called `get_uuid`.
- Refactor and rename the `__fix_field_date` method to `__cast_str_to_datetime` to convert str to datetime and fix possible date format errors. With this method we reduce the number of calls of the `str_to_datetime` method.
- In the `__add_commit_meta_fields` method it only creates a new identity when it doesn't exist in SortingHat.
- In the `add_identity` method, remove the `update_profile` mutation; this is no longer necessary as we are using GraphQL.

### DIFF
--- a/releases/unreleased/enrichment-process-performance-improved.yml
+++ b/releases/unreleased/enrichment-process-performance-improved.yml
@@ -1,0 +1,12 @@
+---
+title: Enrichment processing time reduced by 50%
+category: performance
+author: Quan Zhou <quan@bitergia.com>
+issue: null
+notes: |
+  The general performance was improved reducing the number of calls
+  to the identities manager (i.g. `SortingHat`). There were some deprecated
+  calls that weren't needed any longer and also, we increased the cache of
+  individuals in ELKs.
+  We were also able to reduce the processing time of the Git backend by
+  converting commit dates only once.
--- a/grimoire_elk/elk.py
+++ b/grimoire_elk/elk.py
@@ -371,7 +371,7 @@ def load_identities(ocean_backend, enrich_backend):
     return identities_count
 
 
-@lru_cache(1024)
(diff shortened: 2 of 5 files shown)


releases/unreleased/enrichment-process-performance-improved.yml
```diff
@@ -0,0 +1,12 @@
...
+category: performance
+author: Quan Zhou <quan@bitergia.com>
+issue: null
+notes: |
+  Reduced number of queries to get SortingHat info, reduced number
+  of requests to get backend name and string to datetime conversion,
+  authors in commit message (approved_by, author_by, etc.) only they
+  will be created if the individual does not exist in SortingHat.
+
+  All this improves the performance of the enrichment process.
```
sduenas, line 12:
```suggestion
  The general performance was improved reducing the number of calls
  to the identities manager (i.g. `SortingHat`). There were some deprecated
  calls that weren't needed any longer and also, we increased the cache of
  individuals in ELKs.
  We were also able to reduce the processing time of the Git backend by
  converting commit dates only once.
```

releases/unreleased/enrichment-process-performance-improved.yml
```diff
@@ -0,0 +1,12 @@
+---
+title: Enrichment process performance improved
```
sduenas, line 2:
```suggestion
title: Enrichment processing time reduced by 50% 
```

sduenas (approved):
LGTM but please, edit the changelog file as suggested.

sduenas (approved):
LGTM

PR #1112, opened 2023-07-17
Improve load identities process performance

jjmerchante:
This PR improves the performance of load identities. process by 30-40%. Instead of inserting the identities one by one, now inserts the identities in batches of 100, reducing the number of queries to SortingHat.


sduenas (commented):
LGTM. Can you add the changelog entry about performance before we merge this?

sduenas (approved):
LGTM

PR #1113, opened 2023-07-21
[enrich] Fix enrich items with multiple enrollments

zhquan:
This commit fixes the date format when there are several enrollments in profiles. Since the number of times the `str_to_datetime` method is executed has been reduced to improve performance the `date` field in `get_item_sh_from_id` and `get_item_sh_meta_fields` must be strings.

Test to enrich a profile with multiple enrollments added.


sduenas (approved):
LGTM

PR #1118, opened 2023-10-05
[github] Fix get_identities method not returning all identities

jjmerchante:
This PR fixes a bug that cause that some identities were not loaded in SortingHat and in OpenSearch were displayed as UNDEFINED.

The method `get_identities` was not returning the identities from `reviews_data`.


zhquan (approved):
LGTM

PR #1119, opened 2023-10-05
[demography] Error in OpenSearch response stops demography study

jjmerchante:
Fix an issue that stops the demography study when OpenSearch occasionally provides a response without some expected fields.


sduenas (approved):
LGTM

PR #1120, opened 2023-10-13
[dependencies] Remove PyMySQL from ELK

jjmerchante:
This code removes the dependency with `pymsql` because it is not longer used with the new version of SortingHat.

Related to https://github.com/chaoss/grimoirelab-elk/pull/856


sduenas (approved):
LGTM

PR #1122, opened 2023-10-18
[enrich] Arguments in wrong order in some backends

jjmerchante:
This PR resolves a bug where the arguments in the `__init__` method of certain enrich backends were incorrectly ordered, or missing.

The original signature of the Enrich backend is:
```
def __init__(self, db_sortinghat=None, json_projects_map=None, db_user='',
             db_password='', db_host='', insecure=True, db_port=None, db_path=None,
             db_ssl=False, db_verify_ssl=True, db_tenant=None):
```

But the backends that overwrite that method had this signature:
```
def __init__(self, db_sortinghat=None, json_projects_map=None, db_user='',
             db_password='', db_host='', db_path=None, db_port=None,
             db_ssl=False, db_verify_ssl=True, db_tenant=None):
```

This change requires another change in sirmordred.

### DIFF
--- a/grimoire_elk/enriched/cocom.py
+++ b/grimoire_elk/enriched/cocom.py
@@ -149,10 +149,10 @@ class CocomEnrich(Enrich):
     metrics = ["ccn", "num_funs", "tokens", "loc", "comments", "blanks"]
 
     def __init__(self, db_sortinghat=None, json_projects_map=None,
-                 db_user='', db_password='', db_host='', db_path=None,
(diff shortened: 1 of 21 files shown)


sduenas (commented):
This is nor a real bug. Classes receive parameters as keywords so the order of them is not a problem. The problem is if the library calls these classes without using keyword parameters and positional ones.
I'm going to reject this PR because is not really needed.

PR #1123, opened 2023-10-18
Individual id incorrectly assigned to identity id

jjmerchante:
Correct a bug that results in enriched items using the individual ID as the identity ID, and vice versa.

This way:
- `author_uuid` = `individual.mk`
- `author_id` = `indentity.uuid`


sduenas (approved):
LGTM

PR #1124, opened 2023-10-18
Define the backend name for SortingHat identities

jjmerchante:
This PR allows to change the backend name for SortingHat identities. The GitHub2 backend is updated from 'github2' to 'github'. This way GitHub2 and GitHub share the same identity in SortingHat.

This change could break existing dashboards because the new github comments ids will be different from the old ones.


sduenas (approved):
LGTM

PR #1125, opened 2023-10-19
[tests] Fix wrong params test

jjmerchante:
The Jira backend now has `--api-token` param and the test to check wrong params is broken. Change the parameter `--api-token` with something unexpected like `--foo-bar`.


sduenas (approved):
LGTM

PR #1132, opened 2024-01-30
[tests] Include the password for SortingHat in tests

jjmerchante:
Include the password for SortingHat in the tests and rewrite the workflow to use the default settings and commands from SortingHat.


sduenas (approved):
LGTM

PR #1133, opened 2024-01-30
[ci/release] Install package before running SortingHat

jjmerchante:
Prior to running SortingHat, it's important to install the package. This change ensures that we can utilize the identical version from the package when deploying SortingHat.


sduenas (approved):
LGTM

PR #1134, opened 2024-02-07
[bug] Incorrect date type in identity enrollment retrieval

jjmerchante:
Mediawiki, Meetup are Jira were failing to retrieve the enrollment for an identity. Because the date was not a string.


sduenas (approved):
LGTM

PR #1136, opened 2024-02-23
[raw/kitsune] Include mapping for data.metadata.value too big

jjmerchante:
This PR includes the mapping for Kitsune to fix an error in inserting Perceval items when the field 'data.metadata.value' is too big.


sduenas (approved):
LGTM

PR #1137, opened 2024-03-11
[kitsune] Update enriched fields

jjmerchante:
This PR allows to execute the demography study for Kitsune backend.

The enriched index now includes a unique identifier for questions and answers (`id`), this was causing an error when trying to autorefresh because `uuid` was missing.

Include the default raw fields in `answers` items that were missing: `metadata__updated_on` , `metadata__timestamp`, `offset`, `origin`, `tag`, and `uuid`.


sduenas (approved):
LGTM

PR #1138, opened 2024-03-15
[kitsune] Handle immense terms for Kitsune

jjmerchante:
This code modifies the raw mapping of Kitsune indexes to handle immense terms found in `data.answers_data.content`, and `data.content`. Thus, the corresponding field is converted to text.


sduenas (approved):
LGTM

PR #1140, opened 2024-04-11
[MBOX] Remove credentials from origin and tag

jjmerchante:
This code redefines the method `_fix_item` for the MBOX raw data to remove credentials that may appear in the origin.


sduenas (approved):
LGTM

PR #1151, opened 2024-05-16
Update Kitsune raw mapping

jjmerchante:
This PR updates the mapping for the raw index of Kitsune. It was not including all the fields because dynamic was set to false in the data attribute.


zhquan (approved):
LGTM

PR #1158, opened 2024-07-31
[git] Fix branches study that assigns wrong branches to commits

jjmerchante:
This PR fixes the Git branches study. There was a bug that didn't reset the list of commits assigned to branches, causing commits to be assigned to the wrong branches.


sduenas (approved):
LGTM

PR #1164, opened 2024-08-13
[enrich] Fix 'roles' attribute error

sduenas:
The 'roles' attribute wasn't available for all the enrichers. The attribute has been added to the main class 'Enrich' so subclasses will have it available even when they don't use it.


zhquan (approved):
LGTM

PR #1165, opened 2024-08-29
[enrich] Avoid "-- UNDEFINED --" values

zhquan:
Avoid `-- UNDEFINED --` values for all SortingHat fields when
Mordred loses connection to the SortingHat server during the
enrichment or autorefresh execution. It will keep the values
​of the `name`, `email`, `id` and `uuid` fields.

### DIFF
--- a/grimoire_elk/enriched/enrich.py
+++ b/grimoire_elk/enriched/enrich.py
@@ -826,13 +826,16 @@ class Enrich(ElasticItems):
                                          email=iden['email'],
                                          name=iden['name'],
                                          username=iden['username'])
+        iden['uuid'] = identity_id
+        iden['id'] = identity_id
+        iden['enrollments'] = []
 
         try:
             individual = self.get_entity(identity_id)
             if not individual:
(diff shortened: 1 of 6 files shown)


grimoire_elk/enriched/enrich.py
```diff
@@ -842,9 +845,10 @@ def get_sh_item_from_identity_cache(self, identity_tuple, backend_name):
                 msg = "Identity {} not found in individual returned by SortingHat.".format(identity)
                 logger.error(msg)
                 return sh_item
-        except SortingHatClientError:
+        except SortingHatClientError as e:
             msg = "None Identity found {}, identity: {}".format(backend_name, identity)
-            logger.debug(msg)
+            logger.error(msg)
+            raise SortingHatClientError(e)
             return sh_item
```
jjmerchante, line 852: This line is not required, because it raises the exception in the line before
```suggestion
```

jjmerchante (changes requested):
LGTM It works as expected, Just one change.

PR #1166, opened 2024-08-30
[git] Improve branches study for consistency

jjmerchante:
Previously, while the branches study was running, the `branches` field remained empty or partially filled until the study was completed, leading to incorrect data being displayed on the dashboard.
With this change, the study is first created in an auxiliary field `branches_aux`, and once complete, the results are moved to the `branches` field.


sduenas (changes requested):
LGTM but please fix the issue with flake8.

sduenas (approved):
LGTM

PR #1167, opened 2024-08-30
[elk] Handle OpenSearch page info on pagination

zhquan:
This code allows to handle the page info returned
when iterating over paginated results in OpenSearch.


jjmerchante (approved):
LGTM

PR #1170, opened 2024-10-08
Add keywords for bugzilla backends

sduenas:
This PR adds the keywords field to the enriched indices for the  bugzilla backends.

### DIFF
--- a/grimoire_elk/enriched/bugzilla.py
+++ b/grimoire_elk/enriched/bugzilla.py
@@ -153,6 +153,12 @@ class BugzillaEnrich(Enrich):
         if 'votes' in item['data']:
             eitem['votes'] = item['data']['votes'][0]['__text__']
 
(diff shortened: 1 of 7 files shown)


vchrombie:
Hi @sduenas, 

Would it be helpful to also update the schema to reflect this new field? I understand there’s a lot of pending work with schemas, but it might streamline understanding and usage, even if it’s one step at a time.

What are your thoughts on this?

sduenas:
> Hi @sduenas,
> 
> Would it be helpful to also update the schema to reflect this new field? I understand there’s a lot of pending work with schemas, but it might streamline understanding and usage, even if it’s one step at a time.
> 
> What are your thoughts on this?

Good catch! I've just added it.

jjmerchante (approved):
LGTM

PR #1171, opened 2024-10-10
Add reponse times for bugzilla indices

sduenas:
The Bugzilla enriched items include two new fields to track response times on these type of repositories:

- `time_to_first_attention` is the the time expressed in days between the ticket creation and the first comment from a contributor different from the author creating the bug.
- `last_comment_date` is the date of the last comment posted in the bug.


jjmerchante (approved):
LGTM

PR #1182, opened 2025-04-08
Revert updating dependencies from last release

jjmerchante:
There was a mistake in the Perceval release that caused version 2.x to be created. This commit reverts the poetry.lock to downgrade the Perceval version that was incorrectly released.


sduenas (approved):
LGTM

PR #1184, opened 2025-05-20
Use Perceval backend-specific from_date as default value

jjmerchante:
This PR fixes an issue that stopped GitHub from collecting issues incrementally. The comparison for `from_date` was being made against the date `1970-01-01` instead of using the backend's default `from_date`.


sduenas (approved):
LGTM

PR #1186, opened 2025-06-02
Fix incorrect from-date enriching behavior

jjmerchante:
Remove enriching behavior using `from-date` parameter because it should only apply to collection, not enriching.

There was a bug where `from-date` was being compared against `1970-01-01`, which is incorrect since each backend has its own default `from-date`. This caused enrich tasks to enrich all items based on a shared default value instead of their backend-specific value.

Additionally, using `from-date` value from configuration was causing items collected after that date to be enriched, rather than enriching only the items that were updated in the origin data source since that date.

Removing that code fixes an issue that was causing all items from GitHub to be enriched instead of enriching only those items that are inserted after `metadata__timestamp`.

### DIFF
--- a/grimoire_elk/enriched/utils.py
+++ b/grimoire_elk/enriched/utils.py
@@ -182,29 +182,10 @@ def get_last_enrich(backend_cmd, enrich_backend, filter_raw=None):
 
             filters_.append(filter_raw_)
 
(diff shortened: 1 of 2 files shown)


sduenas (approved):
LGTM

PR #1190, opened 2025-06-19
Module `pkg_resources` is deprecated in favor of `importlib`

jjmerchante:
> Draft Pull Request until we deprecate Python 3.9 because doesn't use the same syntax as in Python >= 3.10

Use of `pkg_resources` is deprecated in favor of `importlib.resources` and `importlib.metadata` since Python 3.12.

Refactor resource loading to use their alternatives.

More info: https://setuptools.pypa.io/en/latest/pkg_resources.html


jjmerchante:
Superseded by https://github.com/chaoss/grimoirelab-elk/pull/1198

PR #1191, opened 2025-06-23
[enriched/github] Fix assignee fields

zhquan:
This commit fixes the `KeyError` when the issue was assigned to a deleted user.

A deleted user does not have these fields:
- email
- company
- location


jjmerchante (approved):
LGTM

PR #1197, opened 2025-08-27
[github/raw] Handle immense term for data.auto_merge.commit_message

zhquan:
This code modifies the raw mapping of `github` indexes to handle immense terms found in `data.auto_merge.commit_message`. Thus, the corresponding field is converted to text.

tests/data/github2.json updated accordingly.


sduenas (approved):
LGTM

PR #1198, opened 2025-08-29
Drop Python 3.9 and ElasticSearch 6 support

jjmerchante:
Update workflows and configuration to support Python 3.10.

Python 3.9 will no longer receive updates after October 2025. This change ensures compatibility with newer features and maintains security for the project.

Remove `elasticsearch` package and use the latest version of `opensearch-py` for future development. This change makes that ElasticSearch 6 is no longer compatible.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,28 +17,44 @@ jobs:
           artifact-path: dist
 
   tests:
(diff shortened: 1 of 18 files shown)


sduenas (approved):
LGTM

PR #1201, opened 2025-09-10
Reduce the number of jobs in workflows

jjmerchante:
This PR removes Python 3.13 from the tests and release workflows to prevent creating a large number of jobs for each test.


sduenas (approved):
LGTM

PR #1202, opened 2025-09-15
[ci] Update workflows to use django-admin runserver

jjmerchante:
Changed the command in the release and tests workflows to use `django-admin runserver`.

This change is made because sometimes the tests fails using the command sortinghatd.


sduenas (approved):
LGTM

PR #1203, opened 2025-09-17
Sync AOC and Git raw indexes error

jjmerchante:
An error occurred in the Areas of Code study, where syncing the AOC and the Git raw index caused an infinite loop due to using different dates for filtering


sduenas (approved):
LGTM

PR #1205, opened 2025-09-19
Add block_size parameter to AreasOfCode study

jjmerchante:
This change introduces a `block_size` parameter to the Areas of code study. It allows for better control over data processing in chunks, and allow to define a lower number when the items to process are big.


sduenas (approved):
LGTM

PR #1208, opened 2025-10-09
[github] Update mapping to disable indexing of raw text fields

jjmerchante:
Update `commit_message`, `body`, and `diff_hunk` fields to non-indexed objects to improve performance and reduce storage for raw GitHub data.


sduenas (approved):
LGTM

PR #1210, opened 2025-10-27
[git aoc/onion] Implement custom bulk write for OpenSearch

jjmerchante:
The current `bulk` helper from OpenSearch tries to decode the documents as UTF-8, and in many cases, there are items that are not compatible and raise an exception.

Add a new `bulk_write` method for Ceres `ESConnector` to handle document uploads to OpenSearch using the bulk API. This new method manages bulk sizes and handles errors during the upload process.


sduenas (approved):
LGTM

PR #1212, opened 2025-11-21
[raw] Use latest date for incremental collection

jjmerchante:
Updated the logic for determining the last update date. Now it considers both `from_date` from `setup.cfg` and the last update from Elasticsearch, ensuring to use the most recent date.


grimoire_elk/raw/elastic.py
```diff
@@ -175,11 +175,11 @@ def feed(self, from_date=None, from_offset=None, category=None, branches=None,
 
         last_update = None
         if 'from_date' in signature.parameters:
-            if from_date:
-                last_update = from_date
+            last_update_es = self.get_last_update_from_es(filters_=filters_)
```
canasdiaz: @jjmerchante What is the value of `last_update_es` if the index is empty? If this is controlled, the changes LGTM.

canasdiaz (approved):

PR #1214, opened 2025-12-04
[git/aoc] Improve bulk write response handling in ESConnector

jjmerchante:
Refactor how bulk operation responses are processed in the ESConnector class. This change fixes errors that occurred when some items in a bulk request failed, ensuring partial failures are handled correctly and do not break the processing loop.


zhquan (approved):
LGTM

PR #1216, opened 2025-12-10
[enrich/bugzillarest] Enrich comments

sduenas:
Bug comments are enriched an included in the index. The new field 'type' contains whether the item is a 'bug' or a 'comment'.

Indices using bugzillarest backend will have to be regenerated to support this new feature.


jjmerchante (approved):
LGTM

PR #1217, opened 2025-12-16
Update index naming in Onion study

jjmerchante:
This change updates the output index name for Onion study to include the current date. This prevents the removal of the previous index until the new one is ready.

This ensures that data is available both during the study and in case it fails.


zhquan (approved):
LGTM

PR #1221, opened 2026-01-20
[bugzilla/enrich] Update Bugzilla ID to be unique

jjmerchante:
This change modifies how IDs are formatted in the BugzillaREST backend.
The ID now includes a prefix to ensure uniqueness across different types of items. This helps avoid conflicts and makes it easier to identify the source of each ID.


sduenas (approved):
LGTM

PR #1222, opened 2026-01-23
[bugzillarest] Fix autorefresh and identity handling for comments

jjmerchante:
Fixes issues where comment roles were missing from the roles attribute, comment identities were not loaded into SortingHat, and `_user_name` fields were removed during autorefresh.


sduenas (approved):
LGTM

PR #1223, opened 2026-01-24
ci: add OpenSSF Scorecard GitHub Action workflow

### LINKED ISSUE #1220: Calculate the OSSF Scorecard
Please add the OpenSSF github action  as explained in https://scorecard.dev/#using-the-github-action, to calculate the Scorecard.

Drishtipixiee:
Fixes #1220 

This PR adds the OpenSSF Scorecard GitHub Action as recommended in the official documentation. The workflow runs on pushes to main, weekly on a schedule, and can be triggered manually. Results are uploaded to the GitHub Security tab and stored as artifacts to continuously monitor the project's security posture.


vchrombie:
Thanks for taking this.

Before we merge this, I think it’s worth stepping back and validating whether adding OpenSSF Scorecard is something we want to adopt more broadly for GrimoireLab from a security and maintenance perspective.

If we decide this is useful, ideally it should be applied consistently across all GrimoireLab repositories rather than only grimoirelab-elk, so we avoid fragmentation and uneven security signals.

I think we can move the issue to the chaoss/grimoirelab repository too.

Drishtipixiee:
I understand your point. enabling OpenSSF Scorecard across all GrimoireLab repositories would be more consistent and useful than introducing it in just one repository. will be glad to follow the discussion in the chaoss/grimoirelab repository and proceed based on the broader decision there.

PR #1239, opened 2026-06-02
[phabricator/enrich] Support user data from the new Phabricator API

jjmerchante:
This PR adds support for enriching Phabricator user data using the new `user.search` API endpoint, in addition to the existing `user.query` endpoint.


sduenas (approved):
LGTM

